PR #1, opened 2019-11-14
Updates in App, css and album

natalieschnelzer:
Didn't have time to look into the hover properly. At the moment, the image fades away, the artist and name get underlined but the icons on top of the image don't appear.

Link to my deployed project:

### DIFF
--- a/code/src/Album.js
+++ b/code/src/Album.js
@@ -1,0 +1,11 @@
+import React from 'react';
+
+export const Album = (props) => {
+    console.log(props)
+    
+ return  <div className="albumCard">
+    <img src={props.album.images[1].url} alt=""/>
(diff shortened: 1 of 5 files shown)


code/src/Album.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+export const Album = (props) => {
+    console.log(props)
+    
+ return  <div className="albumCard">
+    <img src={props.album.images[1].url} alt=""/>
```
JennieDalgren: here you are selecting the second image in the array, is that on purpose?

natalieschnelzer: Yes, there was an option of three different sizes and I chose the middle one.

PR #2, opened 2019-11-22
Project 9 - music site

nyblad:
https://nyblad-music-releases.netlify.com/

Had a hard time to understand how to type the components in App.js:  
`<Album key={item.id} item={item} />`
Took a day or two before I really understood how that worked and how to work with naming and the connections between the names in App.js and what I wrote in Album.js. But now I think I kind of understand how to do. Was good to do the stretch-goal and kind of repeat everything.

- How to change `<Album key={item.id} item={item} />` to be separated something like `<Album key={item.id} image={item.xx} title={item.xx} artist=[item.xx}/>` (don't want to change my codebut just understand how to)

- In `.albumCards` I have a `height: 30%` that fixed the height just as I want it, but I don't get what it is 30 % of..?

- How to structure the code: What shpuld be a separate component and not? What should be in App.js?

- Should classnames be lowercase-dash or camelCase? Like className="album-title" or "albumTitle"?


code/src/components/album.css
```diff
@@ -0,0 +1,72 @@
...
+  margin: 10px;
+  width: 90%;
+  height: 30%;
+  position: relative;
+}
+
+.album-cover {
+  width: 100%;
+  max-height: 300px;
+}
```
ericaglimsholt: This needs some more work in mobile. There the images stretch which is not so nice. Try using `object-fit` here

code/src/components/album.css
```diff
@@ -0,0 +1,72 @@
...
+  transition: all .3s ease-in-out;
+}
+
+.album-title {
+ font-size: 14px;
+ font-weight: bold;
+ color: #fff;
+ font-family: Helvetica;
+ margin-top: 10px;
+ text-decoration: none;
```
ericaglimsholt: Indent

PR #6, opened 2019-11-24
Ninas music site

ninamansster:
Link to my deployed project:
https://ninamansster-music-react.netlify.com/

I started out really low in confidence as I felt bit small as a developer when I compiled the things in the Portfolio. And then I knew that React is hard to grasp, so I felt sort of stressed. However, now when I see the Spotify app looking so nice and that it actually works I am proud.

I am also very proud of my team, that openly shares hard things and shares solutions.

How to select all the objects in an array for example all the artists: 
artistname={album.artists[0].name}?
I also struggled with the icons outside SRC, so I copied the icons-folder to SRC, to be able to call them in article.js. Maybe there is another way?


code/src/App.js
```diff
@@ -1,12 +1,33 @@
...
+      <h1>New Albums and Releases</h1>
+      <div className="appcontainer">
+        {data.albums.items.map((album) => (
+          <Album
+            key={album.id}
+            image={album.images[1].url}
+            uri={album.external_urls.spotify}
+            artistname={album.artists[0].name}
+            artisturi={album.artists[0].external_urls.spotify}
+            albumname={album.name} />
```
JennieDalgren: stick to camelCase for the prop names

JennieDalgren: Regarding the artists.
Pass along the whole artists-array as a prop, so instead of `artistname` and `artisturi`go for `artists={alnum.artists}`

code/src/Album.js
```diff
@@ -0,0 +1,29 @@
...
+
+      <div className="iconscontainer">
+        <div className="icon"><Heart /></div>
+        <a className="icon" href={props.uri}><Play /></a>
+        <div className="icon"><Dots /></div>
+      </div>
+    </div>
+    <div className="artistcontainer">
+      <a className="album" href={props.uri}>{props.albumname}</a>
+      <a className="artist" href={props.artisturi}>{props.artistname}</a>
```
JennieDalgren: Here, to be able to show multiple artists (if there are multiple artists on the album) do a similar solution as you did in App.js to get all the albums. 
```
{props.artist.map(artist => (
))}
```

ninamansster:
Tons of thnks :-)

I fixed it with the help of your instructions, so now I have all the artists (if there are several) with links together with Album.

Deployed again :-)

All the best,
Nina

PR #8, opened 2019-11-24
Frida Jonsson music releases

annafridamaria:
[Spotify releases](https://frijon-music-releases.netlify.com/)

It felt natural splitting up the code into components and styling them separatly, and it was fun styling towards a set design and functionality. I also learnt more about positioning and the z-index.

I still have a hard time really understanding the infratstructure of the information, like how a component that is not actually referring directly to the json can find the information with props. This also makes it hard as I can't really figure out how I should write if I want to map a specific key in an component, for example map the artists for each album. For now I just went with the first artist for every album


code/src/AlbumCover.css
```diff
@@ -0,0 +1,57 @@
...
+  
+ .Icons svg {
+     width: 30px;
+     fill: white;
+  }
+
+  .Icons a {
+    width: 30px;
+    fill: white;
+ }
```
JennieDalgren: do you really have an a tag?

code/src/AlbumCover.js
```diff
@@ -0,0 +1,21 @@
...
+import { Dots} from 'icons/dots'
+import 'AlbumCover.css'
+
+export const AlbumCover = (props) => (
+  <div className="AlbumImage">
+    <div className="Icons">
+      <Heart />
+      <div className="Play">
+        <Play />
+      </div>
```
JennieDalgren: You can put the className on the Play tag instead.
`<Play className="play" />`
also, don't start the classnames with capital letter.

code/src/AlbumCover.js
```diff
@@ -0,0 +1,21 @@
...
+      <Dots />
+    </div>
+
+    {/* Album cover */}
+    <img src={props.item.images[1].url} alt="Album cover"></img>
```
JennieDalgren: Since this is the only data this component needs to know about, it's not necessary to pass in more than that.
See explanation above ☝🏼

code/src/Album.js
```diff
@@ -0,0 +1,17 @@
...
+import { AlbumCover } from 'AlbumCover'
+import { AlbumInfo } from 'AlbumInfo'
+import 'Album.css'
+
+export const Album = (props) => (
+    <div className="Album">
+
+    {/* Album cover */}
+    <AlbumCover 
+     item={props.item}/>
```
JennieDalgren: Since the AlbumCover component only needs to get information about the actual image, you can choose to only pass that in as a prop instead of the whole item object.
Something like this:
```
 <AlbumCover image={props.item.images[1].url}/>
```

code/src/AlbumInfo.js
```diff
@@ -0,0 +1,25 @@
...
+    <div className="Title">
+    <a href={props.item.uri}>{props.item.name}</a>
+    </div>
+    <div className="Artist">
+    {props.item.artists.map((artist) => (
+      <span className="AnArtist">
+      <a key={artist.id}
+      href={artist.uri}>
+        {artist.name}</a>
+        </span>
```
JennieDalgren: perfect, to map through all artists to print them out.

PR #9, opened 2019-11-24
Music releases project by Agnes

Agnes94:
Link to my deployed project: https://eager-dijkstra-e79144.netlify.com/

My reflections on how this week's project turned out: I spent a lot of time on the portfolio project and started with this week pretty late so I already felt stressed because of that. I was feeling very overwhelmed at the beginning and found it very hard to actually get started with this project. The code-along helped a lot and gave me an idea about how to start and as I was doing it I've learned more and more.

Things I'd like to have clarified or explained in more detail: It felt a bit confusing to split my code into chunks and therefore I don't have many components, and put most of my styling in index.css file for instance. I was unable to get the url for the album's website, so I'd appreciate some clarification on that.


code/src/App.js
```diff
@@ -1,12 +1,18 @@
...
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
+    <div className="app">
+      <header>
+        <h1>New albums &amp; singles</h1>
+      </header>
+      <div className="app-components">
```
JennieDalgren: maybe choose a better class name. Such as `albums-container` or `albums-wrapper` or similar.

code/src/Components/Album.js
```diff
@@ -0,0 +1,26 @@
...
+      </div>
+
+      <a className="album-link" href={props.album.external_urls.spotify}>
+        <h2>{props.album.name}</h2>
+      </a>
+      <a className="artist-link" href={props.album.external_urls.spotify}>
+        {props.album.artists.map(artist => (
+          <h3>{artist.name}</h3>
+        ))}
+      </a>
```
JennieDalgren: This should be linked to each artist. not the album.
so move the a into the map.
```
 {props.album.artists.map(artist => (
     <a className="artist-link" href={artist.**theLinkToTheArtist**}>
          <h3>{artist.name}</h3>
     </a>
))}
```

PR #11, opened 2019-11-24
Music releases project

Nasimmhn:
Link to my deployed project:

https://new-music-releases-by-nasim.netlify.com/

I got very confused with the import/export things, and how they are related to each other. And because of that, I didn't make many components. I have watched all the code-alongs a couple of times to grasp  React. It makes sense, but as always it gets confusing as you start writing code. I also struggled with some details in styling, but I got useful hints, both from my team and StackOverflow.

Things I'd like to have clarified or explained:
- As I mentioned, I found it very confusing with chunking the components and also naming them. Do we have to make a component for exempel for header, which only has a 'h1' as its content. When is it worth creating a new component of something?


code/src/Album.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react'
+
+export const Album = (props) => {
+  return (
+    <div className="album">
```
JennieDalgren: even though we're working with React now, you could still use semantic html elements. No stress now, but keep it in mind!

code/src/App.js
```diff
@@ -1,12 +1,30 @@
...
-      Find me in src/app.js!
+      <Header />
+      <section id="album-container">
+        {data.albums.items.map((item) => {
+
+          return <Album key={item.id} albumTitle={item.name}
+            albumURL={item.external_urls.spotify}
+            image={item.images[1].url}
+            playIconURL={item.external_urls.spotify}
+            artists={item.artists} />
```
JennieDalgren: ⭐️

code/src/Album.js
```diff
@@ -0,0 +1,41 @@
...
+          <a href={props.albumURL}>{props.albumTitle}</a>
+        </div>
+        <div className="artist-name">
+
+          {props.artists.map((artist) => {
+
+            return <span key={artist.external_urls.spotify}><a href={artist.external_urls.spotify}>{artist.name}</a><span>,&nbsp;</span></span>
+
+          })}
+
```
JennieDalgren: This is enough, since its just a line of text. But here would be the place where you could've created an Artist component and let that one handle the artist name and url. But it's super small so no need to do it.

PR #12, opened 2019-11-24
My PR for the Music Project

claragyokeres:
Link to my deployed project:
https://musicplayerclara.netlify.com

A stressful week with lots of new information and testing. Still not happy with the end result.

The images and hover.


code/src/App.js
```diff
@@ -1,12 +1,22 @@
-import React from 'react'
-import data from './data.json'
+import React from "react"
+import data from "./data.json"
+import { Album } from "./components/Album"
+import { Artists } from "./components/Artist"
+import { Header } from "./components/Header"
+import { tsPropertySignature } from "@babel/types"
```
JennieDalgren: remove this

code/src/App.js
```diff
@@ -1,12 +1,22 @@
...
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
+    <div className='card'>
+      <Header />
+      <section className='album-container'>
+        {data.albums.items.map((item) => {
+          return <Album key={item.id} albumTitle={item.name} image={item.images[1].url} artists={item.artists[0].name} urlArtistName={item.artists[0].external_urls.spotify} albumUrl={item.external_urls.spotify} />
+        })}
```
Pass along the whole artists-array as a prop, so instead of `artists={item.artists[0].name} urlArtistName={item.artists[0].external_urls.spotify}` go for ` artists={album.artists}`

code/src/components/Album.js
```diff
@@ -0,0 +1,22 @@
...
+            <div className='album-cover'>
+                <img className='album-img' src={props.image} alt='Album' />
+                <Icons />
+            </div>
+            <div className='album-info'>
+                <div className='album-title'>
+                    <a href={props.albumUrl}>{props.albumTitle}</a>
+                    <div className='artist-name'>
+                        <a href={props.urlArtistName}>{props.artists}</a>
+                    </div>
```
JennieDalgren: Here, to be able to show multiple artists (if there are multiple artists on the album) do a similar solution as you did in App.js to get all the albums. 
```
{props.artists.map(artist => (
))}

File: code/src/components/Album.js [hunk 4]
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { Icons } from "components/Icons"
+
+export const Album = (props) => {
+    return (
+        <div className='artistBox'>
```
JennieDalgren: remember, we're working with React but you can still use semantic HTML elements such as section, article and so on. Just to think of in the future.

PR #13, opened 2019-11-25
Music release project week 9 Angelica

Angelicabrodin:
Link to my deployed project:
https://musicreleaseproject.netlify.com

I started pretty late on this project, since I had a hard time finishing with my portfolio project in time. I thought it was pretty hard to get started in the beginning, I felt kinda overwhelmed about all the different files in the directory and what they actually meant. (In the code alongs there were only a few) So I needed a lot of time just understanding the directories and how it actually worked mounting a component and so on. But when I started to understand a bit more I thought it was really fun and it finally felt more logic than it did with vanilla javascript, so this far I prefer react :)
I still struggle a lot in general and need a lot more than 20 hours a week, but I am really proud that I again managed to finish the project and that I am really happy with the result!

- I had a few struggles during this project. for instance I was not able to fetch all the different artists from the array, I know there are more than one artists in a few of the albums, but I just managed to get one, so I really would like som help explaining this. 

- If I had some more time, I would have liked to try adding opacity over the album when hovering over, I am struggling a bit with opacity in general, so maybe some pointers on how to do this?

- I also should have maybe used another component, since I now only use App.js and Album.js to structure it a bit more, and also tried to use map() over the whole data, now it looks a bit messy in App.js maybe?

- Making the svg into components, I tried this, and finally got it to work, but then I really had a hard time styling them and trying to get the hover effect, so I finally gave in and changed them into react components instead, I still dont understand the diffrence and when or why you should use either one or the other, but hopefully I will understand this further along :)


code/src/App.js
```diff
@@ -1,12 +1,26 @@
...
+      <div className="headerSection">
+        <Header />
+      </div>
+      {data.albums.items.map((item) => (
+        <Album
+          key={item.id}
+          albumTitle={item.name}
+          urlArtistName={item.artists[0].external_urls.spotify}
+          image={item.images[1].url}
+          artistName={item.artists[0].name}
```
Pass along the whole artists-array as a prop, so instead of urlArtistName and artistName go for artists={album.artists}

code/src/App.js
```diff
@@ -1,12 +1,26 @@
...
+import './App.css'
 
 console.log(data)
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
+    <div className="albumComponent">
+      <div className="headerSection">
```
JennieDalgren: No need to wrap `<Header /> ` in a div, you can put a className on that instead.

code/src/components/Album.js
```diff
@@ -0,0 +1,23 @@
...
+        <div className="albumCover" style={{ backgroundImage: `url(${props.image})` }}>
+            <HeartIcon className="heartIcon" />
+            <PlayIcon className="playIcon" />
+            <DotsIcon className="dotsIcon" />
+            {/* <Dots className="dotsIcon" /> */}
+        </div>
+        <div className="moreDetails">
+            <a className="linkAlbumTitle" href={props.urlAlbumTitle}><h1 className="albumTitle">{props.albumTitle}</h1></a>
+            <a className="linkArtistName" href={props.urlArtistName}><h3 className="artistName">{props.artistName}</h3></a>
+        </div>
```
JennieDalgren: Here, to be able to show multiple artists (if there are multiple artists on the album) do a similar solution as you did in App.js to get all the albums. 
```
{props.artist.map(artist => (
))}
```

PR #14, opened 2019-11-25
Project music releases, Michel Elmoznino Laufer

MichelLaufer:
Link to my deployed project:
https://michel-musicproject.netlify.com/

I think that the project gave us a very good practise on React and I think that the project was fun. 
I had some initial struggles when starting with the project on understanding how
to use map on the data.

I think that the code sessions on the webinars on Wednesday and Friday were very good and gave clarification to the project. Maybe I would have liked a short video lecture on using map before starting with the project but I got help on StackOverflow on how to do this.


code/src/index.css
```diff
@@ -11,3 +14,205 @@ code {
...
+  text-align: center;
+}
+
+img {
+  height: 240px;
+  width: 240px;
+}
+
+.menu-container {
+  display: grid;
```
idlefingers: 😎

code/src/sidebar.js
```diff
@@ -0,0 +1,16 @@
...
+
+console.log(data)
+
+export default () => {
+  return (
+    <Menu>
+    {data.playlists.items.map((item) => {
+      return <a><a className="album-link" href={item.external_urls.spotify} target="_blank" rel="noopener noreferrer">
+      {item.name}</a></a>
+    })}
```
idlefingers: I think this would be more readable using implicit returns, and some newlines:

```js
    {data.playlists.items.map((item) => (
      <a className="album-link" href={item.external_urls.spotify} target="_blank" rel="noopener noreferrer">
        {item.name}
      </a>
    ))}
```

Also, why double `<a>` tags? One should be enough, I think ;)

One more thing - great work with the target blank technique with 'noopener noreferrer' - great practice! 👏

PR #15, opened 2019-11-25
project-mysuc-releases Susanna Sjöberg

SanSjo:
Link to my deployed project:
https://music-releases.netlify.com/

This weeks project was heard to grasp in the beginning. I made it but I wasn't really sure what I was doing. But in the end of the week it turned out good I think, and i understand a bit more know. Even if its still a bit confusing. More practising I guess :)

I'm not sure when I should divide the different components. How "small" they should be. Now I have Header.js, Albums.js and Icons.js, maybe I should have diveded Albums into more components? 

I'm not really sure yet how many different components I should have in a project.


code/public/index.html
```diff
@@ -1,9 +1,9 @@
 <!DOCTYPE html>
 <html lang="en">
-
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <link rel="stylesheet" src="./src/index.css" />
```
idlefingers: This is unnecessary (and won't actually work when the project is deployed) since the css is being imported in `src/index.js` - it magically becomes part of the build which netlify deploys.

code/src/Albums.js
```diff
@@ -0,0 +1,65 @@
...
+    >
+      <div className="ArtistContainer">
+        <div>
+          <div className="iconsContainer">
+            <Icons />
+          </div>
+          <div className="ArtistContent">
+            <a className="titleLink" href={props.albumURL}>
+              <p className="AlbumName">{props.name}</p>
+            </a>
```
idlefingers: I think the `<p>` should be outside the `<a>` tag here (instead of the other way around) since it makes the link become a block which means if you hover by the side of the artist name, it hovers the artist name - if that makes sense?

It's not _wrong_ to do it this way, but I think the other way would be nicer in this case.

code/src/Albums.js
```diff
@@ -0,0 +1,65 @@
...
+              <p className="ArtistName">{props.artist}</p>
+            </a>
+          </div>
+        </div>
+      </div>
+    </div>
+  );
+};
+
+/*
```
idlefingers: Don't forget to remove your commented out code when you're cleaning up!

code/src/App.js
```diff
@@ -1,12 +1,27 @@
...
+console.log(data);
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
-  )
-}
+    <header>
```
idlefingers: Should everything be in `<header>`?

code/src/Icons.js
```diff
@@ -0,0 +1,31 @@
...
+
+export const Icons = props => (
+  <div>
+    <div className="icons">
+      <div className="iconsContainer">
+        <img
+          className="iconHeart"
+          src="./icons/heart.svg"
+          alt="Heart icon for favourite"
+        ></img>
```
idlefingers: Image tags are self-closing and also, when using images in the public folder, use paths starting with `/`.. so this should be:

```
<img
  className="iconHeart"
  src="/icons/heart.svg"
  alt="Heart icon for favourite" />
```

code/src/Albums.js
```diff
@@ -0,0 +1,65 @@
...
+
+export const Albums = props => {
+  return (
+    <div
+      className="AlbumContainer"
+      style={{ backgroundImage: `url(${props.image})` }}
+    >
+      <div className="ArtistContainer">
+        <div>
+          <div className="iconsContainer">
```
idlefingers: You have a lot of divs and containers here (and in general).. There's nothing specifically wrong with that, but in future projects try to challenge yourself to use fewer divs like this - you can often get away with using very few elements

PR #17, opened 2019-11-25
PR - Tiago Ferreira

tavferreira:
[https://tavferreira-project-music-releases.netlify.com](https://tavferreira-project-music-releases.netlify.com)

Haven't had a lot of time to do the project, but after re-watching Wednesday and Friday sessions everything became clearer. 
The project didn't seem too complicated, but I've realized that I was understanding some React concepts while doing.

Not sure. As mentioned I didn't have much time, so it felt that it was the best I could do. 
I guess I would have some more opinions if I would have dug deeper in the project.
One thing that I can mention is some configurations around React / Node.js that were preventing me to do a better job, 
like having to restart the server every time I made a change. I understand it would be difficult to have an explanation for every case,
and I guess I'm paying the price of being a minority (by using Linux).


code/src/App.js
```diff
@@ -1,12 +1,33 @@
 import React from 'react'
+import { Albums } from './components/Albums.js'
+import { Playlist } from './components/Playlists.js'
```
idlefingers: It's completely fine to use `.js` in imports, but generally, best practice is to exclude it. So just `./components/Playlists` :)

code/src/components/AlbumData.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import { Artists } from './Artists.js'
+import './albumData.css'
+
+export const AlbumData = (props) => {
+    return (
```
idlefingers: I'm being super picky here, but don't forget to check your indentation settings in your editor - this file has ended up with 4 spaces (should be 2). See [this question on stack overflow](https://stackoverflow.com/c/technigo/questions/253) for a bunch of options to check.

code/src/components/Albums.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { AlbumCover } from './AlbumCover.js'
+import { AlbumData } from './AlbumData.js'
+import './album.css'
+
+export const Albums = (props) => {
+    return (
+        <div className="albumCard">
+            <AlbumCover
+                key={props.album.images[0].url}
```
idlefingers: You don't need a key here (or below). You only need keys when you're in a `.map()`

code/src/components/Playlists.js
```diff
@@ -0,0 +1,18 @@
...
+import { PlaylistCover } from './PlaylistCover.js'
+import { PlaylistData } from './PlaylistData.js'
+import './playlists.css'
+
+export const Playlist = (props) => {
+    console.log(props)
+    return (
+        <div className="playlistCard">
+            <PlaylistCover
+                key={props.playlist.images[0].url}
```
idlefingers: Same here with keys

PR #18, opened 2019-11-25
Project-music-releases - Matilda Arvidsson

HIPPIEKICK:
[https://matilda-arvidsson-project-music-releases.netlify.com/](https://matilda-arvidsson-project-music-releases.netlify.com/)

It turned out OK after what feels like 100 hours of work. It was a really hard project for me, I didn't know how to start and I'm still not a 100% sure of what I've actually done. My code feels like a total mess right now, but the actual site looks quite OK in my opinion.

- I still have trouble understanding the relationship between node.js, npm and react.
- I don't fully understand the meaning of key props, are they necessary on all levels?
- How to make my code less messy
- I changed back and forth between div with background-image and regular img-src for the album covers, but I couldn't figure out how to get rid of opacity on the svg-icons when I had the cover as a div with background-image... Is it even possible?


code/public/index.html
```diff
@@ -13,13 +13,13 @@
...
-  </head>
+  <title>Technigo React App</title>
+</head>
 
-  <body>
-    <noscript>You need to enable JavaScript to run this app.</noscript>
-    <div id="root"></div>
-    <!--
+<body>
+  <!-- <script src="../src/index.js"></script>-->
```
idlefingers: Don't forget to remove commented out code.

You don't need to link the javascript - the dev server deals with that when you're working on your computer (after running `npm start`) and when you deploy, netlify runs `npm run build` which spits out a bundled project all nicely linked together.

code/src/App.js
```diff
@@ -1,12 +1,32 @@
 import React from 'react'
 import data from './data.json'
+import { Header } from 'components/Header.js'
```
idlefingers: It's completely fine to use `.js` in imports, but generally, best practice is to exclude it. So just `components/Header` :)

code/src/Cards.js
```diff
@@ -0,0 +1,29 @@
...
+import { Artists } from './components/Artists.js'
+import { Albumcovers } from './components/Albumcovers.js'
+import 'Cards.css'
+import './components/Albumcovers.css'
+import './components/Artists.css'
+
+export const Cards = (props) => (
+  <div className="Cards">
+    <Albumcovers
+      key={props.id}
```
idlefingers: You don't need a key here. Keys are only needed inside `.map`

code/src/Cards.js
```diff
@@ -0,0 +1,29 @@
...
+    />
+    <Albums
+      key={props.id}
+      albumName={props.albumName}
+      albumUrl={props.albumUrl} />
+
+    <div className="artistName">
+      {props.item.artists.map(item => {
+        return (
+          <Artists key={item.id} item={item} />
```
idlefingers: So this is when you need a key :)

PR #19, opened 2019-11-25
Music project-Anna Almestål

anna-ip:
Link to my deployed project:
[https://music-site-by-anna-almestal.netlify.com/]

It's been super hard but also a lot of fun to start with React, I think or hope that I sort of got a hang of it. I've struggled a lot with the hovering and my media queries. 

I left the comments on the @media queries be, cause this is something I still haven't been able to figure out, it's working, but not the way I belive it should, but I can't get it to work, so therefore I think the code is a bit messy cause the @media queries are not where they are supposed to be.


code/src/Album.js
```diff
@@ -0,0 +1,35 @@
...
+
+
+
+/*console.log(data);*/
+/*console.log(data.albums.items[0])*/
+
+
+export const Album = (props) => {
+    return (
+        <div className="album">
```
idlefingers: Looks like you're editor is set up to do 4 spaces for indentation (should be 2). See [this question on stack overflow](https://stackoverflow.com/c/technigo/questions/253) for a bunch of things to check.

code/src/App.js
```diff
@@ -1,12 +1,33 @@
...
+
+      </div>
+      <section className="album-container">
+        {data.albums.items.map(item => {
+          return <Album
+            key={item.id}
+            image={item.images[1].url}
+            name={item.name} artists={item.artists[0].name}
+            link={item.external_urls.spotify}
+            artists_link={item.artists[0].external_urls.spotify} />
```
idlefingers: Props should be written in camelcase, so this last one should be `artistsLink={item.artists[0].external_urls.spotify}`

code/src/Header.js
```diff
@@ -0,0 +1,17 @@
...
+
+export const Header = () => {
+    return (
+        <div className="Header">
+            <div>
+
+                <h1>New albums and singles</h1>
+            </div>
+
+
```
idlefingers: Try to avoid lots of empty lines in your code - it makes it harder to read.

code/src/App.js
```diff
@@ -1,12 +1,33 @@
 import React from 'react'
 import data from './data.json'
+import { Album } from 'Album'
+import { artists } from './artists'
```
idlefingers: Don't forget to remove unused imports when you're tidying up.

code/src/artists.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import data from './data.json';
+
+export const artists = () => {
+    return (
+        <div>
+            <data />
```
idlefingers: This wouldn't work (React has no idea how to render json), but it looks like this component isn't being used anyway, so it should be removed! Less code is always better! :)

PR #20, opened 2019-11-25
music-app-cislowski.netlify.com

KarolCislowski:
Link to my deployed project:
music-app-cislowski.netlify.com

It was a great amount of fun!!


code/src/components/News/Album/Album.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+
+const Album = (props) => {
+  const artistsList = props.artistName.map((element) => {
+    return (<span>{element.name} </span>)
+  })
```
idlefingers: Nitpicking, but this would be nicer as an implicit return. There's nothing wrong with this approach (preparing the elements outside of render and then just using the variable further down), but in larger apps this can make things harder to follow. I'd push for doing the .map within the returning JSX as much as possible, until you hit a point where pulling it up and out is warranted.

code/src/components/News/Album/Album.js
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <article className="album">
+      <img src={props.img} alt="album cover" className="album-img" />
```
idlefingers: Since you have the album name as a prop here, I think that would've made a great alt text :)

code/src/components/News/News.js
```diff
@@ -0,0 +1,36 @@
+import React, { Component } from 'react'
+import Headline from '../Shared/Headline/Headline'
+import Album from './Album/Album'
+import data from '../../data.json'
```
idlefingers: Nothing wrong with relative imports, but just fyi, the project is set up so you can do absolute imports (from 'src'). So these could be:

```
import Headline from 'Shared/Headline/Headline'
import Album from 'components/Album/Album'
import data from 'data.json'
```

It just makes it nice not to have to think so much about folder structure.

code/src/components/News/News.js
```diff
@@ -0,0 +1,36 @@
...
+
+class News extends Component {
```
idlefingers: Function components, please!

code/src/components/News/News.js
```diff
@@ -0,0 +1,36 @@
...
+  constructor() {
+    super()
+    this.state = {
+      news: data.albums.items,
+      title: 'New albums & singles'
+    }
```
idlefingers: Great that you were experimenting with state before we introduced it, but if I'm gonna be picky, these things needn't be in state since they never change.. State is a place for things which change - for everything else, you can use normal variables.

code/src/components/SideBar/SideBar.js
```diff
@@ -0,0 +1,35 @@
+import React, { Component } from 'react'
+import Headline from '../Shared/Headline/Headline'
+import Playlist from './Playlist/Playlist'
+import data from '../../stretch-goal.json'
+
+class SideBar extends Component {
```
idlefingers: Same here with function components! (and the use of state below)

code/src/index.css
```diff
@@ -1,13 +1,146 @@
...
+  opacity: 0.5;
 }
+
+.play:hover {
+  opacity: 1;
+  transform: scale(1.1);
+  -webkit-transform: scale(1.1);
+  -moz-transform: scale(1.1);
+  -ms-transform: scale(1.1);
+  -o-transform: scale(1.1);
```
idlefingers: You don't need vendor prefixes on `transform` - it's supported in all browsers.

Also, the tool which is bundling the css for us actually  uses [autoprefixer](https://www.npmjs.com/package/autoprefixer) so you never need to worry about vendor prefixes when writing css :)

PR #21, opened 2019-11-25
Music Project- Jenny Jacobsson

jennyjacobsson:
https://jenny-project-9-music.netlify.com/

For being my first React project ever, I'm proud to say the site works and looks fine😎. This week was both challenging and fun. I enjoy splitting up the code into smaller components, a much nicer way to work I think. 

I still find it a bit hard to write functions and calling API:s, especially with the arrays and props. I got a lot of help from my mentor with that part. And I guess practice makes perfect 😅

I'm a bit confused on how small or big a component are suppose to be? Is it up to me to decide? Some components in my project are very small, but I thought of it as practise mostly, to export and import components.

Have a  great day ☀️


code/src/Album.js
```diff
@@ -0,0 +1,29 @@
...
+        <Image title={props.name} images={props.images} />
+        <Controls playUrl={props.url} />
+      </AlbumArt>
+
+      <h2 className="album-name">
+        <a href={props.url}>{props.name}</a>
+      </h2>
+      {props.artists.map((artist, i) => {
+        return (
+          <React.Fragment key={artist.id}>
```
idlefingers: Nice with fragments! 😎

code/src/Album.js
```diff
@@ -0,0 +1,29 @@
...
+            {i > 0 && ", "}
```
idlefingers: Nice solution. I think CSS would've been a better place for this type of presentation, though. You could use the `:after` pseudo-element to add a comma, then use `:last-child` to set the content on the final element to be '' (so no comma)

PR #22, opened 2019-11-25
Music releases - Petra Brinck 

petrabrinck:
Link to my deployed project: https://brave-jones-f321b1.netlify.com

It was challenging to get all the details in place but I finally saw some ”clusters" of code and began to get a little better understanding when it comes to React. I tend to make things more complicated than they could be so in the future I will try to find easier solutions with less code.  I forgot mobile first so despite help today I still can’t get media queries to work the way I want them to. I will continue to try to solve it :-) 
I do think It was a great project though. Very inspiring and relevant since we went to an meetup at Spotify and got to meet and listen to several female speakers (all working there) talk about their daily job and their journeys towards becoming developers. 

How do I go about doing media queries when I forget to do mobile first?
Thanks! 
Petra


code/src/Album.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import './components/album.css';
+import './components/icons.css';
+/* import data from './data.json'*/
+
+/*console.log(data)
+console.log(data.album.items[1])*/
+
+export const Album = (props) => {
+    return (
```
idlefingers: Looks like your editor is set up to do 4 spaces for indentation (should be 2). See [this question on stack overflow](https://stackoverflow.com/c/technigo/questions/253) for a bunch of things to check.

code/src/Album.js
```diff
@@ -0,0 +1,48 @@
...
+                </div>
+            </div>
+        </div>
+    )
+}
+
+
+
+
+
```
idlefingers: Try to avoid having lots of blank lines in your code - just to keep things tidy!

code/src/Artists.js
```diff
@@ -0,0 +1,12 @@
...
+
+export const Artists = () => {
+    return (
+        <div>
+            <data />
+            Export Artists!
+        </div>
+    )
+}
+
```
idlefingers: Looks like this file isn't used (and react doesn't know how to render json, so it'd break if it was!). Don't forget to remove unused code.

code/src/index.html
```diff
@@ -0,0 +1,31 @@
+<!DOCTYPE html>
```
idlefingers: This file isn't needed - the one in the public folder is the one which is used.

petrabrinck:
Thank you ever so much Damien! 
I will go thru all comments carefully and update (if not soon, I'll do it over Christmas :-)

PR #23, opened 2019-11-25
Anna Terelius project 9

AnnaTerelius:
Link to my deployed project:
https://annatereliusproject9-music-releases.netlify.com/

I'm happy with the result even tough it was superhard to grasp for example props. I still don't know how to make the album-img stay darker when hovering the Icons section; happy for some help here! :)

well.. props, when to have them and from where to get them..


code/src/Album.js
```diff
@@ -0,0 +1,34 @@
...
+                    <img className="play" src="icons/play.svg" alt="playIcon"/>
+                    <img className="dots" src="icons/dots.svg" alt="dotsIcon"/> 
+                </div> 
+            </div>            
+            <div className="album-title">
+                {props.name}
+            </div> 
+            <div className="artist">
+            {props.artists.map(artist => {   
+                return <Artist name={artist.name} href={artist.href} />
```
idlefingers: Don't forget to use a key here. Keys are only needed within a `.map()`

code/src/Album.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import {Artist} from './Artist'
+import './album.css'
+import './header.css'
+
+export const Album = (props) => {
+    console.log("props", props)
+    return (
```
idlefingers: Looks like your editor is set up to do 4 spaces for indentation (should be 2). See [this question on stack overflow](https://stackoverflow.com/c/technigo/questions/253) for a bunch of things to check.

code/src/Artist.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+import './artist.css'
+
+export const Artist = (props) => {
+    console.log("artist props", props)
+        return (
+            <a className="artist-name" href={props.href}>{props.name}</a>
+        )
```
idlefingers: Keep an eye on your indentation! These 3 lines should be one level less indented

PR #24, opened 2019-11-25
Jola Chabik - Project Music Releases

ucha3i:
Link to my deployed project:

https://ucha3i-project-8-music-app.netlify.com

I was a bit sick last week, and I got behind with this project, unfortunately. I tried to catch up with everything on weekend, and I managed to do basic part of assignment, with a lot of help from Stack Overflow and Slack channel. 

It's still not perfect, and it's not exactly done, but I'd like to hand it in and move to next assignment, and also close second sprint in the meantime. I'd rather not to get even more behind, and that's why I'd rather finish this assignment when I'll do everything for week 10 and for second sprint. 

I guess I got basic level of understanding how is it like to work in React. I'm sure, I'll be able to try it out also in the week 10 task and few times more :) 

I've got a big trouble with Netlify and Github. First, files on Github haven't changed their names (at the beginning I had named them starting from small letter, and changed it later), after that for some reason a few files haven't updated itself, despite a few commits I did, and it was really stressful. 

I also had problems with hover effect, and it still doesn't work as it supposed to. And the last - I wasn't able to make a link from the play icon. I just don't know how to do it, and I'd probably figure it out finally, but I'd like to hand the assignment in. 

I hope it understandable what I wrote here :)  

Thanks :)


code/src/App.js
```diff
@@ -1,12 +1,37 @@
...
+    <div key={id} className="artist-container">
+      <Album 
+      image={images[1].url} 
+      name={name} 
+      url={external_urls.spotify} 
+      />
+      {album.artists.map(renderArtist)}
+    </div>
+  );
+};
```
idlefingers: There's nothing wrong with this approach (making a function to render an element and then invoking it in a map), but in larger apps this can make things harder to follow because you have to jump around a lot to read whats going on. I'd push for doing the .map within the returning JSX as much as possible, until you hit a point where pulling it up and out is warranted.

code/src/App.js
```diff
@@ -1,12 +1,37 @@
 import React from 'react'
 import data from './data.json'
+import './app.css'
+import { Artist } from "./components/Artist";
+import { Album } from "./components/Album";
 
-console.log(data)
+const renderAlbum = album => {
+  const { id, images, name, external_urls } = album;
```
idlefingers: Awesome that you're using destructuring! You could go one step further and remove a little more code by moving up to where you handle the function arguments:

```js
const renderAlbum = ({ id, images, name, external_urls }) => {
  // ...
}
```

code/src/components/Album.js
```diff
@@ -0,0 +1,24 @@
...
+import { Icons } from './Icons'
+
+export const Album = props => {
+  const { image, url, name } = props;
+  return (
+    <div className="poster">
+      <div className="album-cover">
+        </div>  
+      
+    <div className="release">
```
idlefingers: Keep an eye on your indentation here. It makes it really hard to see if there's an error in your html with an unclosed element or not.

PR #27, opened 2019-11-25
Music-site-Aleksandra

AleksandraMuller:
Hello!

Link to my deployed project:
https://music-site-aleksandra.netlify.com/

The start was very rough. The first two days, I spent on learning about React, Components, did extensive research and Google was really my friend. At the beginning I was a bit unsure on how to start, but after getting to know React a bit, I understood the concept of Components and props. I started to code my project from plain JS, then I split some parts of that into components. I made a loop inside of another loop, and found that bit tricky. I am pretty sure there is an easier way to fix this. My mentor then helped me during a coding lunch to pack some of my code into functions, to make it look cleaner. We also used destructuring which also made code looks simpler. Lastly, I worked on Icons which were a struggle, but I made them work. Hover effect over the play button was suprisingly the biggest challenge. Finally, I based my hover solutions on the one that Karol had. So I made an box with icons that lays on the top of the image.

Overall, I am happy that I managed to achieve.

In my opinion the site needs quite some more improvements in case of responsiveness, naming the classes, components, the JSX structure of the content and grid. My goal is to change that as soon as I will get some more time. 

Can we make the loop extracted and use it in another Component with props, and later extract that logic into App.js?

Thank you!


code/src/Components/Artist/Artist.js
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+
+export const Artist = props => {
+  return (
+    <a href={props.url} target="_blank" rel="noopener noreferrer">
+      {props.name}
+    </a>
```
JennieDalgren: one little thing here is that when there is more than one artist they are shown flush together, like `Teyana TaylorKehlani` instead of `Teyana Taylor Kehlani`

code/src/App.js
```diff
@@ -1,12 +1,30 @@
...
+
+import { Artist } from "./Components/Artist/Artist";
+import { Album } from "./Components/Album/Album";
+
+import data from "./data.json";
+
+const renderArtist = artist => {
+  const { id, name, external_urls } = artist;
+  return <Artist key={id} name={name} url={external_urls.spotify} />;
+};
```
JennieDalgren: nice done creating functions. 
there's no need to do it this way, you could do it all in the return. But if you feel this is clearer, it's all good.

PR #32, opened 2019-11-27
project-music-releases

jensnilsen:
Link to my deployed project:
https://fervent-wright-cf5cac.netlify.com/

Im verry happy about the result. struggled a bit with react and working with components but feel like i get it. At least better and better. 
Been struggling over all with school at the moment. But i'll get there. 

I'm not sure. the hardest part was to write components. Understanding what to write in what component. I feel like i just need to put more hours into it.


code/src/Album.css
```diff
@@ -0,0 +1,89 @@
...
+  margin: 0 20px 0 20px;
+  fill: white;
+  background: rgba(233, 233, 233, 0.07);
+  border-radius: 50%;
+  opacity: 0.9;
+}
+
+.play:hover {
+  transform: scale(1.1);
+  transition: all 0.2s ease-in-out;
```
JennieDalgren: nice! 😎

code/src/Album.css
```diff
@@ -0,0 +1,89 @@
...
+  right: 0;
+  top: 0;
+  left: 0;
+  background: rgba(0, 0, 0, 0.5);
+}
+
+.theLogos svg:hover {
+  background: none;
+  opacity: 1;
+}
```
JennieDalgren: great job with the hover

code/src/Album/Album.js
```diff
@@ -0,0 +1,30 @@
...
+    <div className="albumDetails">
+      <a className="albumName" href={props.item.external_urls.spotify}>
+        {props.item.name}
+      </a>
+      <a
+        className="artistName"
+        href={props.item.artists[0].external_urls.spotify}
+      >
+        {props.item.artists[0].name}
+      </a>
```
JennieDalgren: Here, to be able to show multiple artists (if there are multiple artists on the album) do a similar solution as you did in App.js to get all the albums. 
```
{props.artist.map(artist => (
))}
```

code/src/App.js
```diff
@@ -1,12 +1,28 @@
...
+        <Header />
+      </div>
+      <div className="albumList">
+        {data.albums.items.map(item => (
+          <Album
+            key={item.id}
+            item={item}
+            albumURL={item.external_urls.spotify}
+            artists={item.artists[0].name}
+            artistURL={item.artists[0].external_urls.spotify}
```
Pass along the whole artists-array as a prop, so instead of artists and artistURL go for artists={album.artists}

code/src/Header.js
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+
+export const Header = () => (
+  <div>
+    <p className="headertext">New Albums &amp; singles</p>
+  </div>
```
JennieDalgren: use semantic html. <header> instead of <div>

PR #33, opened 2019-11-28
Music Release Project. 

swolfson1:
Link to my deployed project:
https://project-9-music.netlify.com

it was hard to grasp in the beginning, i also realized that my i can become blind to my own code very easily i have trouble seeing small errors that other people see. i need take time to "spell check"  my code.
i could spend a whole other week trying to make it perfect but i realized i needed to step away from it. 
THE ICONS! and the hover i don't understand why it doesn't work for me. 
what in my CSS do i need to change? please help.


code/src/Artist.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+
+export const Artist = (props) => {
+  return (
+    <div className="artist">
+      <a href={props.url} className="artistLink">{props.name}</a>
+    </div>
```
JennieDalgren: take away the div around the a here. That will make the artists go on the same row. right now the divs (one for each artist) are rendered inside of a p.

PR #41, opened 2020-03-11
Music Releases – Kajsa Holmström & Emma Prené

crushoftheyear:
https://music-releases-react.netlify.com/

We created three components; one for album card, one that holds the artist info, and one for the icons overlay. We also added one CSS file for each component. If we had more time we would have added more components (e.g. album cover) to get a better structure.


code/src/components/AlbumCard.js
```diff
@@ -0,0 +1,33 @@
...
+import { HoverOverlay } from './HoverOverlay.js'
+import { Artist } from './Artist.js'
+import './albumcard.css'
+
+export const AlbumCard = (props) => {
+  return (
+    <article className="album-card">
+
+      <div className="album-cover">
+        <img src={props.src} alt=""></img>
```
ericaglimsholt: Maybe the album name would suit good here in the alt?

code/src/components/HoverOverlay.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import './hover.css'
+
+export const HoverOverlay = () => {
+  return (
+
+    <div className='overlay'>
```
ericaglimsholt: You are using `"` over `'` when it comes to the markup, so it would be nice if this follows the same rule
```suggestion 
    <div className="overlay">
```

PR #51, opened 2020-03-13
Project Music Releases Johanna R & Johanna Å

jastrand:
Link to my deployed project: 
https://infallible-sammet-2dbca6.netlify.com/

We manage to start up the project early and learned by discussing our different issues. We are happy with how this project turned out. 

If the album name is long, the artist names are printed at different levels in the desktop view. We tried many techniques to solve this, but did not manage in the end. We would like some advice on this. We also wonder if css "fill: white" is only possible when converting svg files to components? We only managed to make them white with another solution using filter: invert.


code/src/Album.css
```diff
@@ -0,0 +1,60 @@
...
+  width: 300px;
+  height: 300px;
+  display: flex;
+  justify-content: space-evenly;
+  position: absolute;
+  background: black;
+  opacity: 0.75;
+}
+
+.Img-wrapper {
```
ericaglimsholt: How come these to starts with big letters when the rest is starting with a small letter? :)

code/src/Albums.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import { Artist } from './Artists'
+import { HoverIcons } from 'Hovericons'
+
+
+// This compontent will fetch all data related to the Album including Album name and album image.
```
ericaglimsholt: Nice comment! Very useful

code/src/Albums.js
```diff
@@ -0,0 +1,24 @@
...
+export const Album = (props) => {
+  return (
+    <article className='album-card'>
+      <div className="Img-wrapper">
```
ericaglimsholt: You seem to use `'` everywhere else so it's good if you follow the same here (and check my comment before this one about the big letter)
```suggestion
      <div className='img-wrapper'>
```

code/src/Albums.js
```diff
@@ -0,0 +1,24 @@
...
+        <HoverIcons />
+        <img src={props.item.images[1].url} alt='album cover' />
```
ericaglimsholt: Suggestion: having the album title as `alt`

code/src/Hovericons.css
```diff
@@ -0,0 +1,23 @@
+
```
ericaglimsholt: This file needs some work on the indention

code/src/Album.css
jastrand: I believe it was a rookie mistake on this one - would you say that small letters are best to work with using when declaring classes?

code/src/Albums.js
jastrand: Great idea!

PR #65, opened 2020-10-15
Project music releases: Olof, Therese & Rebeka

rebkova:
Link to our deployed project: https://app.netlify.com/sites/threol-music-release/deploys


code/src/Album.js
```diff
@@ -0,0 +1,47 @@
...
+  //create array of objects containing artist name, URL for each album.
+  const artistsArr = props.artists.map((artist) => {
+    const name = artist.name;
+    const url = artist.external_urls.spotify;
+    return { name, url };
+  });
+
+
+  return (
+    <>
```
EmOlsz, line 19: If your component returns only one root element, there is no need to wrap it around React Fragment.

```
return (
    <div>My element 1</div>
    <div>My element 2</div>
);
```

```
return (
    <>
        <div>My element 1</div>
	<div>My element 2</div>
    </>
);
```

```
return (
    <div>My element 1</div>
);
```

code/src/Album.js
```diff
@@ -0,0 +1,47 @@
...
+              <DotButton className="album-icon dots-icon" />
+            </div>
+          </div>
+          <h2>
+            {props.name}
+          </h2>
+        </a>
+        {/* Mount artist component for each element in the artists array */}
+        {artistsArr.map((artist, index, arr) => (
+          <Artist
```
EmOlsz, line 36: Well done :)

code/src/App.js
```diff
@@ -1,12 +1,28 @@
...
+export const App = () =>
+  (
+    <div className="master-container">
+      <h1 className="site-header">New albums &#38; singles</h1>
+      <div className="header-line"></div>
+      <div className="albums">
+        {albumArr.map((album) =>
+
+          //Create one album component for each item in the album array
+          <Album
```
EmOlsz, line 17: I'm surprised this worked, you should wrap this component in parentheses (as in line 35)

code/src/Artist.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+export const Artist = (props) => {
+
+  //determine if artist should have comma after or not
+  if (props.length === 1)
```
EmOlsz, line 6: Great job!

code/src/index.css
```diff
@@ -1,13 +1,188 @@
...
+}
+
+.artist-name {
+  font-family: Helvetica;
+  color: #a0a0a0;
+  font-size: 14px;
+}
+
+.artist-name:hover {
+  color: #ffffff;
```
EmOlsz, line 103: `color: #fff`

code/src/index.css
```diff
@@ -1,13 +1,188 @@
...
+  img {
+    width: 200px;
+  }
+}
+
+.icon-container {
+  position: absolute;
+  top: 0;
+  right: 0;
+  left: 0;
```
EmOlsz, line 156: Thank you for proper order!

EmOlsz (approved):
First review this week and it already looks great :)
Clean, consistent code. Do you also feel that it's so easy to keep your code more clean with React? That was my first impression when I started working with this library.
Take a look on some comments I left and try to remember about them :)

code/src/App.js
rebkova: Hmm ... I think the Album component does have parenthesis - in line 14 and 26. It's wrapped inside the `map` function. We omitted `{}` and `return` after `=>`  since the Album component is the only thing we're returning from the` map()`. Let me know, if I misunderstood you.

code/src/index.css
rebkova: :) I think CSS returns 6 letters by default, so we just let it :honeybee:

code/src/App.js
EmOlsz: I checked it out and your flow is proper, my bad Rebeka :)

PR #69, opened 2020-10-15
Anna Hasselström & Claire Caudwell Music release project

ClaireCaudwell:
https://objective-wing-b8f314.netlify.app/

1.  Good practice to start working with React to then be able to use this on the next project when we will be working with React. 
2. Was complex, but we found way's of working to achieve the end goal. 
3. Felt that it would've been good to have the lectures based specifically on this project, rather than on a similar project e.g using the same names and accessing the information from the data.json.

1. Relationship between index.js and app.js e.g. if you always need to use the word App as the first component to get the project up and running.
2. How NPM works and the relationship between it and also the react libraries. 
3. Folder structure when working with React and components.


code/src/App.js
```diff
@@ -1,12 +1,18 @@
-import React from 'react'
-import data from './data.json'
+import React from 'react';
+import data from './data.json';
```
EmOlsz, line 2: There is a convention which says that we should divide outer dependencies (those located in node_modules) from local dependencies (all local files that we created - components, functions, data such as arrays) by one empty line. Moreover, `style.css` should be separated as well and located below local dependencies. Ideally, it should look like this :

```jsx
import React from 'react';

import './style.css';
```

So, final order is :

1. Outer dependencies
2. Local dependencies
3. Style file(s)

code/src/App.js
```diff
@@ -1,12 +1,18 @@
...
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
-  )
-}
+    <main>
+      <header>
```
EmOlsz, line 10: At the end I see you didn't make it with header in the same line as main content of app :(
If you are more curious about that, I suggest you to ask Stack Overflow question and tag me ;)

code/src/App.js
```diff
@@ -1,12 +1,18 @@
...
+        <h1>New albums & singles</h1>
+      </header>
+      {/* Below accessing the items array that's inside the album object and sending it 
+      as the prop items_array to Musiclist */}
+      <MusicList item_array={data.albums.items} />
```
EmOlsz, line 15: Props names should be written in camelCase, so rather `itemArray`

code/src/card-component-directory/MusicArtist.js
```diff
@@ -0,0 +1,42 @@
...
+    if (array.length -1 === index) {
+        separator =  "";
+    } else if (array.length -2 === index) {
+        separator = " & ";
+    } else {
+        separator = ", ";
+    }
+
+    return (
+        <span key={artist.id} className="separator"><a className="artist-name" href={artist.external_urls.spotify} target="_blank" rel="noopener noreferrer">{artist.name}</a>{separator}</span>
```
EmOlsz, line 30: Great job with separators!
One small thing - this line of code is far to long. Consider line breaks for each prop

EmOlsz (approved):
Good job Anna & Claire, I'm happy to see such well done project of yours :)
I want to say that in general application is more than good, some additional comments were left by me, so feel free to check them out. Consider them as a chance to improve quality of your code even more!

PR #72, opened 2020-10-16
Music Project - Lili & Ellen

ellenbjo:
Link to my deployed project:
https://peaceful-ride-2526af.netlify.app/

Read about reflections in the read me :)


code/src/App.js
```diff
@@ -1,12 +1,26 @@
 import React from 'react'
 import data from './data.json'
-
-console.log(data)
+import './index.css'
```
EmOlsz, line 3: There is a convention which says that we should divide outer dependencies (those located in node_modules) from local dependencies (all local files that we created - components, functions, data such as arrays) by one empty line. Moreover, `style.css` should be separated as well and located below local dependencies. Ideally, it should look like this :

```jsx
import React from 'react';

import './style.css';
```

So, final order is :

1. Outer dependencies
2. Local dependencies
3. Style file(s)

code/src/App.js
```diff
@@ -1,12 +1,26 @@
...
-      Find me in src/app.js!
-    </div>
+    <>
+      <Header />
+      <section className="album-container">
+        {data.albums.items.map((album) => {
+          return (
+            <Album 
+            key = {album.id}
+            albumTitle = {album.name}
```
EmOlsz, line 16: Indentation of attributes could be improved a bit over here

code/src/components/Album.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import './Album.css'
+
+export const Album = (props) => {
```
EmOlsz, line 4: This component could be divided to many smaller components. If there are more than 20 JSX lines of code then it's a clear information you could probably do it better

code/src/components/Album.js
```diff
@@ -0,0 +1,46 @@
...
+            </a>
+
+            <a href={props.albumLink} target="_blank" rel="noopener noreferrer">
+              <h2>{props.albumTitle}</h2>
+            </a>
+
+            <div>
+              {props.artists.map((artist) => {
+                return (
+                <a
```
EmOlsz, line 29: Wrong indentation

EmOlsz (approved):
Proper project ladies, good job :) As I mentioned in comments, it could be done in a more clean, divided way, but it's not that big of a deal since it's super small. For future projects however I would like you to pay extra attention to that!

PR #81, opened 2020-10-16
project-music-releases - Karin & Emma 

emma-tech:
**Music releases pair-programming project with @karinnordkvist**

**Link to our deployed project:**
https://spotify-by-emma-karin.netlify.app/

**Our reflections on how this week's project turned out:**
We're very happy with the project :) React is hard to grasp in the beginning, but it was good that we got the opportunity to work on this as a team. Karin is more experienced as a developer and could quicker grasp the technical parts, and was then able to help me (Emma) understand more about the things that I thought was hard 👍 
We worked really well together as a team, but maybe could've planned our time a little better by breaking down the project into smaller parts and work a little bit separately as well, and then merge the pieces and work on the bigger picture together. 

**Things we'd like to have clarified or explained in more detail:**
How to move data between components. How to modify our components a bit, using props to be able to map through the data within the component.


code/src/App.js
```diff
@@ -1,12 +1,24 @@
...
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
-  )
-}
+    <main className="main-wrapper">
+      <div className="all-content--wrapper">
```
EmOlsz, line 10: Great job with BEM :)

code/src/Components/Album.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import Icons from 'Components/Icons';
```
EmOlsz, line 2: There is a convention which says that we should divide outer dependencies (those located in node_modules) from local dependencies (all local files that we created - components, functions, data such as arrays) by one empty line. Moreover, `style.css` should be separated as well and located below local dependencies. Ideally, it should look like this :

```jsx
import React from 'react';

import './style.css';
```

So, final order is :

1. Outer dependencies
2. Local dependencies
3. Style file(s)

code/src/Components/Artist.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+
+const Artist = (props) => {
+  return (
+    <>
```
EmOlsz, line 5: You don't need this fragment element, because you are returning only one element inside (being precise, this element inside is another fragment, but there it's ok because deeper we have anchor and span :) )

code/src/Components/Artist.js
```diff
@@ -0,0 +1,34 @@
...
+            rel="noopener noreferrer"
+            className="artist-name"
+          >
+            {props.name}
+          </a>
+
+          <span className="artist-text">
+            {props.amountOfArtists > 2 &&
+            props.index >= 0 &&
+            props.index < props.amountOfArtists - 2
```
EmOlsz, line 22: Clean, consistent approach, good job!

code/src/index.css
```diff
@@ -1,13 +1,203 @@
...
+
+/* ----- Album-section ----- */
+.all-albums--wrapper {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
+}
+
+.album--wrapper {
+  width: calc(25% - 10px);
```
EmOlsz, line 57: Great that's the best way to center elements!

code/src/index.css
```diff
@@ -1,13 +1,203 @@
...
+  list-style-type: none;
+}
+.list--link {
+  color: #a0a0a0;
+}
+
+.list--link:hover {
+  color: #fff;
+  border: none;
+  cursor: default;
```
EmOlsz, line 180: `cursor` property should be added to default element (by default I mean not hovered, so rather on `.link--link` than on `.list--link:hover`).
Moreover, UX standards says that if element is clickable, then cursor should be set to `pointer`

EmOlsz (approved):
Fantastic project ladies! :) It's almost as clean as it could be, moreover you've implemented a lot of extra requirements. I could go like that for a long time : BEM methodology, `calc` properties, clean media queries, extra sidebar. I'm super happy :)
Take a look on few of comments that I left, they might help you improve even more (but only a bit, because you are almost on the top regarding this week :) )

PR #82, opened 2020-10-16
music-release-project - Sara Stjernberg & Evelina Berg & Anna Hellqvist

AnnaHellqvist:
Link to my deployed project:
https://frosty-shirley-67a23d.netlify.app/

It was a fun project, it was a bit difficult to switch to react. It was great to do a pair programming for the first project. 
The hardest part was to add the , and & to the artist links. 

In general how to use if-statements in react.  For this project, how to get the comma and & outside of the links.


code/src/Album-cover-overlay.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import Play from './icons/play.svg';
```
EmOlsz, line 2: There is a convention which says that we should divide outer dependencies (those located in node_modules) from local dependencies (all local files that we created - components, functions, data such as arrays) by one empty line. Moreover, `style.css` should be separated as well and located below local dependencies. Ideally, it should look like this :

```jsx
import React from 'react';

import './style.css';
```

So, final order is :

1. Outer dependencies
2. Local dependencies
3. Style file(s)

code/src/App.js
```diff
@@ -1,12 +1,17 @@
...
-console.log(data)
+import Card from './Card';
+import PageTitle from './Page-title';
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
+    <>
```
EmOlsz, line 7: Proper usage of fragment, good job :)

code/src/Artist-name.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+
+const ArtistName = (props) => {
+  return (
+    props.artistsArray.map((artist, i, array) => {
+      return (
+        <a className="artist-link" key={artist.id} href={artist.external_urls.spotify}>
+          <p className="artist-name">
+            {array.length > 1 && array.length!==2 && i < array.length - 1  && i !== array.length-2  ? artist.name + ',  '
```
EmOlsz, line 10: Your code is working, but is not very readable - consider using normal if statements in such complex situations.
Ternary operators are great, but mostly when used in easy situation like if/else (so, when we have only 2 options)

code/src/Card.js
```diff
@@ -0,0 +1,37 @@
...
+import AlbumName from './Album-name'
+import ArtistName from './Artist-name'
+import data from './data.json';
+import AlbumCoverOverlay from './Album-cover-overlay'
+
+const albumsArray = data.albums.items;
+
+const Card = () => {
+    return (
+        <>
```
EmOlsz, line 12: If your component returns only one root element, there is no need to wrap it around React Fragment.

```jsx
return (
	<div>My element 1</div>
        <div>My element 2</div>
);
```

```jsx
return (
	<>
		<div>My element 1</div>
	        <div>My element 2</div>
	</>
);
```

```jsx
return (
	<div>My element 1</div>
);
```

code/src/Card.js
```diff
@@ -0,0 +1,37 @@
...
+            {albumsArray.map((album) => {
+                return (
+                    <div className="card" key={album.id}>
```
EmOlsz, line 15: `key` attribute is considered as required only in root element inside loop. We don't have to add it anywhere else. And remember - keys should be unique (if there are actual id's in the data - use that)

This snippet of code is wrong :

```jsx
<div key={someUniqueValue}>
  <h1 key={someUniqueValue}>Heading</h1>
  <p key={someUniqueValue}>Paragraph</p>
  <span key={someUniqueValue}>Span</span>
</div>
```

This snippet of code is correct :

```jsx
 <div key={someUniqueValue}>
  <h1>Heading</h1>
  <p>Paragraph</p>
  <span>Span</span>
</div>
```

EmOlsz (changes requested):
Good job ladies! Your code is pretty clean and logic well done. Nonetheless, there are things you need to change - please, take a look on my comments and provide fixes

EmOlsz (approved):

PR #85, opened 2020-10-16
Return of the Spotify Clone - Henrike Wiemker & Karan Mann

karanmann:
**Music Releases**
The project is a simple clone of Spotify showing a list of released albums and singles. I did it in pair programming together with Henrike Wiemker, whose site you can see here: https://github.com/HenrikeW

**The problem**
The clone was built using React and it is the first React-project I ever did. A json-file with data on new releases was provided so the task was to create a React structure using components and passing in parts of the provided data using props. We also fulfilled specific design requirements: When hovering over an album cover, the cover gets darker and three white buttons show up in the middle of it. To achieve this, we used an overlay. If there are several artists who contributet to a release, all of them are shown and devided by a comma.

The site is responsive for mobile, tablet and desktop view and built with a mobile-first approach.

**View it live**
The project is deployed on Netlify and can be viewed live here: https://spotify-music-clone.netlify.app/


code/src/App.js
```diff
@@ -1,12 +1,43 @@
 import React from 'react'
 import data from './data.json'
```
EmOlsz, line 2: There is a convention which says that we should divide outer dependencies (those located in node_modules) from local dependencies (all local files that we created - components, functions, data such as arrays) by one empty line. Moreover, `style.css` should be separated as well and located below local dependencies. Ideally, it should look like this :

```jsx
import React from 'react';

import './style.css';
```

So, final order is :

1. Outer dependencies
2. Local dependencies
3. Style file(s)

code/src/App.js
```diff
@@ -1,12 +1,43 @@
...
+import { ReleaseTitle } from './ReleaseTitle.js'
+import './App.css'
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
+    <div className='section-container'>
+      <Header />
+      <section className='card-container'>
```
EmOlsz, line 13: This part of code could be moved to separate component, don't you think? ;)

code/src/Header.css
```diff
@@ -0,0 +1,9 @@
+.header {
+  font-family: Helvetica;
+  color: White;
```
EmOlsz, line 3: `color: white;` - let's pay attention to lowercase letters!

code/src/ReleaseTitle.css
```diff
@@ -0,0 +1,14 @@
+.release-title {
+  font-family: Helvetica;
+  font-size: 14px;
+  color: #ffffff;
```
EmOlsz, line 4: `color: #fff`

EmOlsz (approved):
Good, solid project, congrats Henrike & Karan :) Some things could be improved (like more components instead of long JSX in one of them), you will find more details in my comments. Take a look on them :)

PR #88, opened 2020-10-17
Project Music-site Emelie Svensson & Sara Heijkenskjöld

emeliesv:
Link to our deployed project:
https://clever-chandrasekhar-21ea25.netlify.app/ 

Our reflections on how this week's project turned out:
This weeks project turned out great! We are very happy with the result and the fantastic teamwork. 
Both of us worked very hard and made it to the black level, so that's very impressive!

Things we'd like to have clarified or explained in more detail:
We interpreted the assignment about responsiveness that the layout with 1 card displayed on mobile, 2 on tablet and 4 on desktop was excluding the playlist, so the playlist is shown all the time. Hopefully this was correct (we like the way it turned out at least).


code/src/App.js
```diff
@@ -1,12 +1,47 @@
...
   return (
-    <div>
-      Find me in src/app.js!
-    </div>
-  )
-}
+    <>
+      <h1 className="app__header">New albums {'&'} singles</h1>
+      <main className="app__grid">
+        <aside className="aside">
```
EmOlsz, line 17: Very nice semantic tag, congrats!

code/src/App.js
```diff
@@ -1,12 +1,47 @@
...
+        </aside>
+
+        <section className="app__item">
+          <h2 className="app__subheader">Singles</h2>
+          {singleArray.map(album => {
+            return <Card key={album.name} album={album} />;
+          })}
+        </section>
+
+        <section className="app__item app__item--album">
```
EmOlsz, line 38: Great that you are using BEM methodology :)

code/src/Artist.js
```diff
@@ -0,0 +1,25 @@
...
+  if (props.index === 0) {
+    return (
+      <a className="artist__name" href={props.artistURL}>
+        {props.artistName}
+      </a>
+    );
+  } else if (props.index === props.length - 1) {
+    return (
+      <a className="artist__name" href={props.artistURL}>
+        {'\u00A0& ' + props.artistName}
```
EmOlsz, line 14: In general great job. Nonetheless, it could be improved - for now, if we hover over the link, both link text and `,` or `&` are underlined. What you could do is wrapping separator mark with another tag, like this :

```
return (
   <>
      <span>\u00A0&</span>
      <a className="artist__name" href={props.artistURL}>
        {props.artistName}
      </a>
   </>
    );
```

code/src/Card.js
```diff
@@ -0,0 +1,44 @@
...
+      <Album
+        key={album.name}
+        albumName={album.name}
+        albumURL={album.external_urls.spotify}
+      />
+      <div className="artist__container">
+        {artistsArray.map((artist, index) => {
+          return (
+            <Artist
+              key={artist.name}
```
EmOlsz, line 31: I'm happy to see you didn't pass `index` as value

code/src/Icons.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import './icons.css';
+
+const Icons = () => {
+  return (
+    <>
```
EmOlsz, line 6: If your component returns only one root element, there is no need to wrap it around React Fragment.

```jsx
return (
    <div>My element 1</div>
    <div>My element 2</div>
);
```

```jsx
return (
    <>
        <div>My element 1</div>
	<div>My element 2</div>
    </>
);
```

```jsx
return (
    <div>My element 1</div>
);
```

EmOlsz (approved):
Great job ladies, truly great job. We had a lot of things to get used to during week 9 and you made it :) Extra congrats for BEM methodology and both red and black requirements.
Always there are things that can be improved, so take a look on my comments. They might help you in future projects!

PR #89, opened 2020-10-17
Sabina and Claudias Spotify Music Clone

Snabbina:
Link to my deployed project:

How to get the artist link to show correctly.....

Warning that showed up in the terminal during the project - what they mean and how to solve them best?
(from1:1)The answer was using props.name though we already destructured the props.

Why cover img behaves different when deployed, disappears on hover instead of getting an overlay 
effect like it hade before deploying? 
(from1:1)The answer was that opacity should have the value of 0.4 instead of 40% to also show correctly on Netlify.


code/src/components/Header.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import "./header.css";
+
+export const Header = () => {
+  return <header className="main-heading">
+    New albums and singles
+      </header>
+};
+
+
```
HIPPIEKICK, line 10: Use parentheses when returning several lines:
```
export const Header = () => {
  return (
    <header className="main-heading">
      New albums and singles
    </header>
  )
};

```
and one newline is enough at the end

code/src/components/AlbumContainer.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import { AlbumCard } from "./AlbumCard";
+
```
HIPPIEKICK, line 3: One empty line is enough

code/src/components/AlbumContainer.js
```diff
@@ -0,0 +1,25 @@
...
+
+export const AlbumContainer = (props) => {
+  const { albums } = props;
+
```
HIPPIEKICK, line 7: One empty line is enough

code/src/components/AlbumContainer.js
```diff
@@ -0,0 +1,25 @@
...
+            artists={album.artists}
+            image={album.images[1].url}
+            albumLink={album.external_urls.spotify}
+            artistLink={album.external_urls.spotify}
+          />
+        );
+      })}
+    </article>
+  );
+};
```
HIPPIEKICK, line 25: Finish off with one newline

code/src/components/AlbumContainer.js
```diff
@@ -0,0 +1,25 @@
...
+    <article className="album-container">
+      {albums.map((album) => {
+        return (
+          <AlbumCard
+            key={album.id}
+            title={album.name}
+            artists={album.artists}
+            image={album.images[1].url}
+            albumLink={album.external_urls.spotify}
+            artistLink={album.external_urls.spotify}
```
HIPPIEKICK, line 19: These two are the same. I'd say it's enough with the `albumLink `and then for the artists' URL, it's enough to have the `artists` so that we can map later on

code/src/components/AlbumCard.js
```diff
@@ -0,0 +1,36 @@
...
+        <Dot className="dot" />
+      </div>
+
+      <a href={albumLink} target="_blank" rel="noopener noreferrer">
+        <h1>{title}</h1>
+      </a>
+      {artists.map((artist, index) => {
+        return (
+          <div key={artist.name} index={index}>
+            <a href={artistLink} target="_blank" rel="noopener noreferrer" >
```
HIPPIEKICK, line 28: This is where you want to specify the path to the artistLink. If I'm not mistaking it should be:
`artist.external_urls.spotify`

netlify.toml
```diff
@@ -3,4 +3,4 @@
 [build]
   base = "code/"
   publish = "build/"
-  command = "npm run build"
+  command = "CI=false && npm run build"
```
HIPPIEKICK, line 6: ⭐
What errors do you get though?

code/src/app.css
```diff
@@ -0,0 +1,78 @@
...
+    grid-template-columns: repeat(2, 1fr);
+  }
+}
+
+@media (min-width: 1024px) {
+  .album-container {
+    display: grid;
+    grid-template-columns: repeat(4, 1fr);
+  }
+}
```
HIPPIEKICK, line 78: The `album-card`s should probably be named `album-card` - because they all contain one card each, right? And they don't fit in the `album-container` in small desktop size so you get a horizontal scroll:
Skärmavbild 2020-10-26 kl  11 28 46

code/src/app.css
```diff
@@ -0,0 +1,78 @@
...
+
+.album-cards {
+  position: relative;
+}
+.image-container {
+  position: relative;
+}
+
+.image-container:hover {
```
HIPPIEKICK, line 22: The hover effects look weird in some widths as well:

Skärmavbild 2020-10-26 kl  11 30 36

HIPPIEKICK (changes requested):
Good job with this project, but please fix;
- responsiveness (hover effects and horizontal scroll)
- correct artists links

netlify.toml
Snabbina: Hade to put that command there to be able to deploy at all, probably due to previous vulnerabilities that prevented us from deploying, since then changes have been made so the issues are resolved now.

code/src/components/AlbumContainer.js
Snabbina: Yes, fixed it now, so the artist url is mapped in the Album card component!

code/src/components/AlbumCard.js
Snabbina: Yes, thats right, fixed it now!

code/src/app.css
```diff
@@ -0,0 +1,131 @@
...
+.play {
+  width: 60px;
+  transition: 0.4s ease;
+}
+.play:hover {
+  transform: scale(1.3);
+}
+
+.album-container {
+  max-width: 1200px;
```
HIPPIEKICK, line 88: I would remove this and either change the media queries a little bit to better suit the widths of your cards, like this. And FYI, since you've already specified that it's `display: grid;` you don't have to specify it again.
```
@media (min-width: 650px) {
  .album-container {
    grid-template-columns: repeat(2, 1fr);
  }
}

@media (min-width: 950px) {
  .album-container {
    grid-template-columns: repeat(3, 1fr);
  }
}

@media (min-width: 1300px) {
  .album-container {
    grid-template-columns: repeat(4, 1fr);
  }
}

```

HIPPIEKICK: Another approach to try out 👇 Don't change the grid, but instead, change the span of each card. For it to be fully responsive you would also need to set the `img`'s width to be no more than `100%` :)

```
.album-container {
  display: grid;
  grid-template-columns: repeat(12, 1fr);
  grid-gap: 1rem;
  position: relative;
  background-size: cover;
  min-height: 50vh;
  padding: 20px;
  margin: 0 auto;
}

.album-card {
  grid-column: span 12;
}

img {
  max-width: 100%;
}

/* Designed from Mobil first, from 668px show 2 cards */
@media (min-width: 668px) {
  .album-card {
    grid-column: span 6;
  }
}

@media (min-width: 1024px) {
  .album-card {
    grid-column: span 3;
  }
}
```

HIPPIEKICK (approved):

code/src/app.css
Snabbina: @HIPPIEKICK I´ve experimentet with the grid and got responsiveness working correctly, yey! But now the icons are behaving weird, I have changed size and padding after the media@ but Im looking for a better solution, could you see what Im missing, I have push my changes.

HIPPIEKICK: Do you mean that they are not fully centered? Your albums now have a dynamic size but your icons have not - so maybe change either the position or the size of the icons to a dynamic width? :)

PR #91, opened 2020-10-18
Lillemor Eriksson & Joel Harde

LillanEriksson:
Link to my deployed project:
https://zen-franklin-39903e.netlify.app/

We are really happy with how the project turned out. We pushed ourselves really hard to reach black level, and we also tried using state for the header dropdown menu. It's not perfect, and we are aware that the menu pops up when the page is loading.

Deploying a react-app to Netlify, it was not an easy ride.


code/src/Albums.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import { Artist } from './Artist';
```
EmOlsz, line 2: There is a convention which says that we should divide outer dependencies (those located in node_modules) from local dependencies (all local files that we created - components, functions, data such as arrays) by one empty line. Moreover, `style.css` should be separated as well and located below local dependencies. Ideally, it should look like this :

```jsx
import React from 'react';

import './style.css';
```

So, final order is :

1. Outer dependencies
2. Local dependencies
3. Style file(s)

code/src/Albums.js
```diff
@@ -0,0 +1,58 @@
...
+import { ReactComponent as PlayButton } from './play.svg';
+import { ReactComponent as HeartButton } from './heart.svg';
+import { ReactComponent as DotButton } from './dots.svg';
+
+export const Albums = (props) => {
+	if (props.type === 'album') {
+		return (
+			<>
```
EmOlsz, line 10: Woooow, some sick indentation :D
Please, for future projects try to pay attention to that :)

Moreover, if your component returns only one root element, there is no need to wrap it around React Fragment.

```jsx
return (
    <div>My element 1</div>
    <div>My element 2</div>
);
```

```jsx
return (
    <>
        <div>My element 1</div>
        <div>My element 2</div>
    </>
);
```

```jsx
return (
    <div>My element 1</div>
);
```

code/src/Albums.js
```diff
@@ -0,0 +1,58 @@
...
+									/>
+								);
+							})}
+						</div>
+					</div>
+				</section>
+			</>
+		);
+	} else {
+		return null;
```
EmOlsz, line 56: 👍

code/src/App.js
```diff
@@ -1,12 +1,62 @@
...
+								albumUrl={item.external_urls.spotify}
+								artistName={item.artists}
+							/>
+						))}
+					</section>
+					<h2>New albums</h2>
+					<section className="main-section">
+						{data.albums.items.map((item) => (
+							<Albums
+								key={item.id}
```
EmOlsz, line 36: I'm happy you used `id` instead of `index` provided by `.map()` method!

code/src/Artist.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+const signsBetween = (index, artists) => {
+	if (index === artists.length - 1) {
```
EmOlsz, line 4: Good job!

code/src/Header.js
```diff
@@ -0,0 +1,48 @@
+import React, { useState } from 'react';
+import {
+	Dropdown,
+	DropdownToggle,
```
EmOlsz, line 4: Extra congratulations for using external packages :) Hope you liked it!

code/src/index.css
```diff
@@ -1,13 +1,298 @@
...
+}
+
+.header-text-container {
+	height: 70px;
+}
+
+.header-text-container h1 {
+	margin-left: 100px;
+	font-size: 40px;
+	margin-top: 0px;
```
EmOlsz, line 25: We don't need to write `px` if value is `0`

code/src/index.css
```diff
@@ -1,13 +1,298 @@
...
+	opacity: 0.5;
+}
+
+.play-button:hover {
+	transform: scale(1.5);
+	opacity: 1;
+}
+
+.album-name {
+	color: #ffffff;
```
EmOlsz, line 174: `color: #fff;`

EmOlsz (approved):
Great job this week! A lot of extra requirements done, external package included, consistent styles present... congratulations :)
Take a look on my comments though, they might help you improve even more!
And please, try to use 2 spaces indentation next time ;)

PR #95, opened 2021-03-11
Project Music Release - Hannah Jesinkey & Jessika Hage

jessika-hage:
Link to my deployed project:


code/src/components/Images.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+// Component for the image and icons on top
+
+export const Images = props => {
```
KarolCislowski, line 5: It looks better when you deconstruct your props here. 
``` 
export const Images = ({ image }) => { }
 ```

File: code/src/components/Header.js [hunk 2]
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+
+export const Header = () => {
```
KarolCislowski, line 3: if you have a component which is only rendering something without doing any logic outside the return block you can use shorthand: 
``` 
export const Header = () => ( 
        <header className ='header'> 
            <h1 className="h1">New albums & singles</h1> 
        </header> 
    )
```

code/src/components/Header.js
```diff
@@ -0,0 +1,9 @@
...
+    return ( 
+        <header className ='header'> 
+            <h1 className="h1">New albums & singles</h1>
```
KarolCislowski, line 6: it is not a good idea to name your classes the same as HTML tags.

code/src/components/Images.js
KarolCislowski, line 5: It is a convention to name your component in plural form when if it contains a collection of other components. 
`Posts` contains a lot of  `Post`. So here it should be `Image`

code/src/components/Playlists.js
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import stretchgoal from "../stretchgoal.json";
+
+// Component for the sidebar with the playlists, includes playlist name and image with play icon on top
+
+export const Playlists = () => {
+    return (
+        <aside className="playlists">
+            <h2 className="playlist-header">Top Playlists</h2>
+            {stretchgoal.playlists.items.map(playlist => {
```
KarolCislowski, line 10: It would be more readable if you use `stretchgoal.playlists.items.map(playlist => ( ) )` to avoid using `return` nested in another `return`

KarolCislowski (approved):
It is an awesome project 🌈 
You need to play a little more with arrow functions to use the easiest format possible for your solution 🦄

PR #96, opened 2021-03-12
Week 9 - project music releases

rasmuselmgren:
This was as group project with myself and Andrea Larsson.

Link to github repo:
https://github.com/rrpersson/project-music-releases

Link to my deployed project:
https://week9-spotify-albums.netlify.app/


code/src/App.js
```diff
@@ -1,12 +1,33 @@
-import React from 'react'
-import data from './data.json'
+import React from "react";
+import data from "./data.json";
+import Album from "./components/Album";
+import Header from "./components/Header";
```

code/src/App.js
```diff
@@ -1,12 +1,33 @@
...
-    <div>
-      Find me in src/app.js!
-    </div>
-  )
-}
+    <>
+      <Header />
+
+      <div className="album-container">
+        {albums.map((album) => {
```
KarolCislowski, line 18: It would be more readable if you use `albums.map((album) => ( ) )` to avoid using return nested in another return

code/src/components/Album.js
```diff
@@ -0,0 +1,54 @@
+import React from "react";
+
+export const Album = (props) => {
```
KarolCislowski, line 3: It looks better when you deconstruct your props here.
```
export const Album = ({ artists, image...... }) => { }
```

code/src/components/Album.js
```diff
@@ -0,0 +1,54 @@
...
+          <a key={artist.id} href={artist.external_urls.spotify}>
+            <span className="album-artist">{artist.name} </span>
+          </a>
+        );
+      })}
+    </div>
+  );
+};
+
+export default Album;
```
KarolCislowski, line 54: You already exported it before.

KarolCislowski (approved):
It is an awesome project 🌈 
You need to play a little more with arrow functions to avoid using nested returns.

PR #97, opened 2021-03-12
Pull request

Sartish:
Link to my deployed project:
https://kind-shaw-7f2831.netlify.app/

Sara Carlstein and Ati Takrami

-it went great!
-Good team work
-Good structure of the assignment


code/src/App.js
```diff
@@ -1,12 +1,25 @@
 import React from 'react'
 import data from './data.json'
+import Album from './components/Album'
+import Header from './components/Header'
+import stretchGoal from './stretch-goal.json'
+import Sidebar from './components/Sidebar'
```
KarolCislowski, line 7: There is a convention that says that we should divide outer dependencies (those located in node_modules) from local dependencies (all local files that we created - components, functions, data such as arrays) by one empty line. Moreover, style.css should be separated as well and located below local dependencies.

code/src/components/Album.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import Artist from './Artist'
+import AlbumImage from './AlbumImage'
+
+
+const Album = (props) =>{
```
KarolCislowski, line 6: If you destructure your props like this:

```jsx
}
```

or like this:

```jsx
  //
}
```

you don't have to write the props parameter when using the props:
`~~<h1>{props.title}</h1>`~~
This is enough:
~~~~`<img src={image} alt="Album cover" />`

code/src/components/Artist.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+
+
+const Artist = (props) => {
+    return(
+        <div>
+            {props.item.artists.map((artist) => {
```
KarolCislowski, line 7: It would be more readable if you use `props.item.artists.map((artist) => ( ) )` to avoid using return nested in another return

KarolCislowski (approved):
It is an awesome project 🌈 
You need to remember to clean all the console logs next time 🦄

PR #98, opened 2021-03-12
Astrid & Pauline 

Pauan8:


code/src/App.js
```diff
@@ -1,12 +1,53 @@
-import React from 'react'
-import data from './data.json'
-
-console.log(data)
+import React from "react";
+import playlist from "./playlist.json";
+import data from "./data.json";
+import { Album } from "./components/Album";
+import { Sidebar } from "./components/Sidebar";
```

code/src/components/Album.js
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+import { MenuBtns } from "./MenuBtns";
+import { Artists } from "./Artists";
+
+export const Album = (props) => {
```
KarolCislowski, line 5: If you destructure your props like this:

```jsx
}
```

or like this:

```jsx
  //
}
```

you don't have to write the props parameter when using the props:
`<h1>{props.title}</h1>`
This is enough:
`Album cover`

code/src/components/Sidebar.js
```diff
@@ -0,0 +1,14 @@
...
+export const Sidebar = (props) => {
+  return (
+    <>
+      <ul>
+        <li>
+          <a href={props.link}>{props.list}</a>
+          <p> Total tracks: {props.tracks}</p>
+        </li>
+      </ul >
+    </>
```
KarolCislowski, line 12: If your component returns only one root element, there is no need to wrap it around React Fragment.

KarolCislowski (approved):
It is an awesome project 🌈 
I really like how you show all the extra data 🦄

PR #99, opened 2021-03-12
project-music-releases: Maria P,  Fethullah D,  Maidelin R

MariaPogosjan:
https://music-releases-now.netlify.app/

It was really fun to start we React. React is easier to understand and the logic is amazing. 

- Key 
- How to use filter method in react


code/src/components/Single.js
```diff
@@ -0,0 +1 @@
+Sing
```
KarolCislowski, line 1: If you resigned to create this component you should remove it.

code/src/components/Artist.js
```diff
@@ -0,0 +1,21 @@
...
+    className="artist-name"
+    href={albumlink}
+    rel="noopener noreferrer"
+    >
+      <>{artistcontainer.map(artist => {
+          return (
+            <p className="artist" key={artist.id}>{artist.name}</p>
+          )
+        })}
+      </>
```
KarolCislowski, line 16: Why you used `<></>` here?

KarolCislowski (approved):
It is an awesome project 🌈

PR #101, opened 2021-03-12
Project Music Releases - Laima & Matilda

leelalaim:
Project was created by Laima and Matilda (Lions team)

Link to my deployed project:
https://music-releases-site.netlify.app/

This weeks project went good. The problems that we faced (regarding adding a commas and and symbols between different artists, aligning, different view on Safari browser..) were sorted with the help of StackOverflow and code coach.

- All of our component are now in the Album.js and then we import the Album component to App.js. is that a correct way of structuring the components or they all should have been in App.js instead of an Album?


code/src/App.js
```diff
@@ -1,12 +1,38 @@
...
+              artist={album.artists.map((item, index) => <a>{item.name}</a>)}
+              hrefArtist={album.artists.map((item) => item.external_urls.spotify)}
+              key={album.id}/>
+             )
+          })
+        }
+      </div>
     </div>
+    </>
   )
```
KarolCislowski, line 35: If your component returns only one root element, there is no need to wrap it around React Fragment.

```jsx
return (
	<div>My element 1</div>
        <div>My element 2</div>
);
```

```jsx
return (
	<>
          <div>My element 1</div>
	  <div>My element 2</div>
	</>
);
```

```jsx
return (
	<div>My element 1</div>
);
```

code/src/components/Icons.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+
+const Icons = (prop) => {
```
KarolCislowski, line 3: You do not use any props so you do not need to specify it here.

code/src/components/AlbumTitle.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+
+const AlbumTitle = (prop) => {
```
KarolCislowski, line 3: If you destructure your props like this:

```jsx
}
```

or like this:

```jsx
  //
}
```

KarolCislowski (approved):
It is a very nice page 🦄

leelalaim:
Requested changes were made

PR #102, opened 2021-03-12
Project 9 Submission Åsa Hildebrand and Shaun Newmark

petercreations:
Link to our deployed project:
https://music-site-project-9.netlify.app/


code/src/App.js
```diff
@@ -1,12 +1,25 @@
-import React from 'react'
-import data from './data.json'
+import React from 'react';
+import data from './data.json';
+import { Album } from './components/Album';
+import { Header } from './components/Header';
+import './index.css';
```

code/src/components/Album.js
```diff
@@ -0,0 +1,37 @@
+  import React from 'react';
+
+  export const Album = (props) => {
```
KarolCislowski, line 3: If you destructure your props like this:

```jsx
}
```

or like this:

```jsx
  //
}

KarolCislowski (approved):
It is a very nice portfolio page 🦄

PR #104, opened 2021-03-12
Project 9 - Isabella Billgren

Bellsibub:
!!See ReadMe for project info!!
Link to published project: https://reacting-to-music.netlify.app


code/src/actions/toggleSidebar.js
```diff
@@ -0,0 +1,7 @@
+// Trigger: User clicks on hamburger menu
+export const toggleSidebar = () => {
+  document.getElementById('sidebar').classList.toggle('active');
+  document.getElementById('overlay').classList.toggle('display');
+};
+
+export default toggleSidebar;
```
KarolCislowski, line 7: Great functionality 🥇

code/src/components/Album/Album.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import moment from 'moment';
+import './Album.css';
+import Artists from 'components/Artists/Artists';
+import AlbumImage from 'components/AlbumImage/AlbumImage';
```
KarolCislowski, line 5: There is a convention that says that we should divide outer dependencies (those located in node_modules) from local dependencies (all local files that we created - components, functions, data such as arrays) by one empty line. Moreover, style.css should be separated as well and located below local dependencies.

code/src/components/AlbumIcons/AlbumIcons.js
```diff
@@ -0,0 +1,23 @@
...
+    <div className="album__icons">
+      <Heart className="icon icon--smaller" />
+      <a
+        className="icon icon--interactable"
+        href={props.url}
+        target="_blank"
+        rel="noreferrer">
+        <Play className="icon--big" />
+      </a>
+      <Dots className="icon icon--smaller" />
```
KarolCislowski, line 18: It would be nice to have a subtle hover effect on your hearts and dots and for sure the cursor: pointer indicating that they are clickable.

code/src/index.css
```diff
@@ -1,13 +1,67 @@
...
+  font-size: 14px;
+  line-height: 1.5em;
+  font-weight: normal;
+  text-decoration: none;
+}
+.text--small {
+  font-size: 0.8em;
+  text-transform: uppercase;
+}
+.list__title {
```
KarolCislowski, line 38: why `__` instead of `--`?? when you are using the BEM you should use `--` to point the element and `__` the state or action. 

`navigation--item__active`

KarolCislowski (approved):
It is an awesome page 🦄

code/src/index.css
Bellsibub: Oh really?! I thought it was the other way around 🙈 
I will take note if this 😉

code/src/components/Album/Album.js
Bellsibub: Great to know! I was actually wondering about this

PR #105, opened 2021-03-12
Project week 9, music releases - Elle Frankenberg & Isabella Markskog 

IsabellaM5:
**Link to our deployed project:**
https://dazzling-aryabhata-b33e50.netlify.app/ 

**Our reflections on how this week's project turned out:**
We did all the blue and red level requirements. 

This was the first week with React, which was a completely new concept for us. In the beginning, we struggled a bit with the different React concepts such as what a component is and how to use props. We tried to read the project brief and analyze how we needed to structure it all, but it wasn't until we had a late night team session that some things fell into place. After that, we got an idea on how to structure the components that enabled us to start coding. 

We're happy with how to project turned out! It was a fun project to start using React and that we started with a pair programming project. 

**Things we'd like to have clarified or explained in more detail:**
Not anything that we can think of at the moment regarding this project. We look forward to next project and dive deeper into the React world! 🐬🐬


code/src/components/Album.js
```diff
@@ -0,0 +1,48 @@
...
+      <div className="album-cover">
+        <a href={props.albumLink}>
+          <img className="album-cover-image" src={props.images[0].url} alt={props.albumName} /> 
+        </a>
+
+          <div className="icons">
+            <img className="icon-heart grow" src="/icons/heart.svg" alt="Heart icon" />
+            <img className="icon-play grow" src="/icons/play.svg" alt="Play icon" />
+            <img className="icon-dots grow" src="/icons/dots.svg" alt="Dots icon" />
+          </div>
```
KarolCislowski, line 26: I can't see them on hover when I'm on 2 or 1 album per row width.

code/src/components/Album.js
```diff
@@ -0,0 +1,48 @@
...
+      </div>
+
+      <a className="album-name" href={props.albumLink}>{props.albumName}</a>
```
KarolCislowski, line 29: Both artist and album title section does not have a hover effect when I'm on 2 or 1 album per row width.

code/src/components/Album.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import { Artist } from './Artist'
+
+export const Album = (props) => {
```
KarolCislowski, line 4: When your album title is longer then 1 line it makes your album photo break the grid on top. Your photos should be in line and your text areas should be ready for longer text. 
image

code/src/components/Album.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import { Artist } from './Artist'
```
KarolCislowski, line 2: There is a convention that says that we should divide outer dependencies (those located in node_modules) from local dependencies (all local files that we created - components, functions, data such as arrays) by one empty line. Moreover, style.css should be separated as well and located below local dependencies.

KarolCislowski (changes requested):
It is a great project. You need to adjust it to the smaller screens and it would be perfect. 🌈

IsabellaM5:
Hey Karol,

We have now made the requested changes to the code, see the recent push to the repo.

We had a follow up question regarding the hover effects. We placed them in the css-file so that they would only apply on the desktop view (breakpoint 1025 px) since we figured that hover effects are only visible on desktop and not mobile or tablet. Is it a convention then to place the css for hover effects on mobile breakpoint even though they are not visible? 😄

/Elle & Isabella 🐬🐬

KarolCislowski (approved):
Great Job 🦄

PR #106, opened 2021-03-12
Our Fancy Music Project (Caroline, Malin, & Elaine)

CoffeeDoodles:
Developers: Caroline Byström, Malin Gustafsson, Elaine Bergström

Link to my deployed project:
https://fancy-music-project.netlify.app/

We found breaking down the project into pieces and pasting it on a organization site like Trello helped up all keep track of where we were in the project. We would probably like to try and get more separated components working the next time but are proud of what we did achieve with our current project. We stuck to our deadline and worked well together. go team!

We would like to discuss more about multiple components with multiple .map functions and just how to smoothly slap them together in the code. We'd also like to know how to substitute an '&' instead of commas for the red level requirements.


code/src/App.js
```diff
@@ -1,12 +1,33 @@
 import React from 'react'
 import data from './data.json'
+import Album from './Components/Album.js'
+import Header from './Components/Header.js'
```

code/src/Components/Album.js
```diff
@@ -0,0 +1,37 @@
...
+
+//image needs an URL
+const Album = (props) => {
+  console.log(props)
+    return (
+      <div className="album-card">
+          <div className="image-container">            
+              <img className="album-image" src={props.imgsrc}></img>          
+            <Buttons />
+          </div>
```
KarolCislowski, line 12: When your album title is longer then 1 line it makes your album photo break the grid on top. Your photos should be in line and your text areas should be ready for longer text.

image

code/src/Components/Album.js
```diff
@@ -0,0 +1,37 @@
...
+            <Buttons />
+          </div>        
+        <div className="album-information-container">
+          <a className="url-link" href={props.url} target="_blank">
+            <h3 className="album-name">{props.name}</h3>
+          </a>
+        </div>
+
+        {props.artist.map((artist) => {
+          console.log(artist)
```
KarolCislowski, line 20: remember to remove console.logs  from your code after you finish debugging your code.

KarolCislowski (changes requested):
This is a very nice project! You need to experiment a little more with your grid to make it align better. 🦄

code/src/App.js
CoffeeDoodles: So we also need to import the style.css into the app.js file?

code/src/Components/Album.js
CoffeeDoodles: I believe we've fixed this issue now! :)

KarolCislowski (approved):
Nice 😸

PR #110, opened 2021-03-13
Malin Vannesjö och Ylva Landoff Lindberg Music Site

sssylvasss:
Link to my deployed project:

https://ylva-malins-amazing-music-site.netlify.app

It was a really good week. The project was difficult enough but not overwhelming.

Where to add what in react. Where to do javascript. I also wanted to understand props better from the beginning, the video was way to short.


code/src/App.js
```diff
@@ -1,12 +1,61 @@
...
-import data from './data.json'
+import React from 'react';
+import data from './data.json';
+import json from './stretch-goal.json';
+import { Header } from './Components/Header.js';
+import { Album } from './Components/Album.js';
+import { Playlists } from './Components/Playlists.js';
+import { Playlistheader } from './Components/Playlistheader.js';
+import { HeaderSingles } from './Components/HeaderSingles.js';
+import { Singles } from './Components/Singles.js';
```
KarolCislowski, line 9: There is a convention that says that we should divide outer dependencies (those located in node_modules) from local dependencies (all local files that we created - components, functions, data such as arrays) by one empty line. Moreover, style.css should be separated as well and located below local dependencies.

code/src/Components/Album.js
```diff
@@ -0,0 +1,35 @@
...
+                    <div className="artist-container">
+                    {
+                        props.artists.map(artist => (
+                            <Artists artistName={artist.name} artistURL={artist.external_urls.spotify}/>
+                        ))
+                    }
+                    </div>
+                </div>
+            </div>
+        </>
```
KarolCislowski, line 33: If your component returns only one root element, there is no need to wrap it around React Fragment.

```jsx
return (
	<div>My element 1</div>
  <div>My element 2</div>
);
```

```jsx
return (
	<>
		<div>My element 1</div>
	  <div>My element 2</div>
	</>
);
```

```jsx
return (
	<div>My element 1</div>
);
```

code/src/Components/Album.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { Artists } from './Artists.js';
+
+export const Album = (props) => {
```
KarolCislowski, line 4: destructure your props like this:

```jsx
}
```

or like this:

```jsx
  //
}
```

code/src/Components/Album.js
```diff
@@ -0,0 +1,35 @@
...
+    console.log(props);
```
KarolCislowski, line 5: remember to remove console.logs  from your code after you finish debugging your code.

KarolCislowski (approved):
It is an avesome page 🦄

PR #111, opened 2021-03-13
New branch and pull request for Music-releases-project by Estefanía Quevedo and Therese Hagelin

esteficodes:
Link to my deployed project: https://cranky-colden-77d5cc.netlify.app/

This week's project I had a very good experience concerning teamwork. My teammate (Therese Hagelin) and I were able to do all the work together by pair-programming and we could also meet up with other group teammates to share progress and exchange advices.
I got a good kickstart on React components and got to finally understand how to work with svg files.

I would like to continue to work on the map() method as I think it's not completely clear to me. And also, I think I could use some more practice with flexbox as i keep using grid because I don't have the handle of it yet.


code/src/Album.js
```diff
@@ -0,0 +1,22 @@
...
+export const Album = (props) => {
+    return (
+        <div className="card"> 
+          <div className="picture">
+            <a href={props.albumurl}>         
+            <img className="image" src={props.picture} alt="album cover"/></a>
+            <div className="icons">
+            <img className="heart-icon" src="icons/heart.svg" alt="" />
+            <img className="play-icon"  src="icons/play.svg" alt="" />
+            <img className="dots-icon"  src="icons/dots.svg" alt="" />
```
KarolCislowski, line 13: Your album picture should be still blurred when you hover over icons.

code/src/App.js
```diff
@@ -1,12 +1,27 @@
 import React from 'react'
 import data from './data.json'
+import {Album} from './Album.js'
+import {Header} from './Header'
+import './index.css'
```

code/src/App.js
```diff
@@ -1,12 +1,27 @@
...
 
 console.log(data)
```
KarolCislowski, line 7: remember to remove console.logs  from your code after you finish debugging your code.

code/src/Header.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+
+export const Header = (props) => {
```
KarolCislowski, line 3: You do not use props inside of your function so it is no need to declare it here.

KarolCislowski (changes requested):
You are very close to passing this assignment! Please work a little more on your hover effects to fix the disappearing blurred effect issue.

PR #112, opened 2021-03-13
Week 9 - Music releases project - Therese Hagelin & Estefanía Quevedo

ThereseHag:
Link to my deployed project:

https://therese-estefania-music-releases-app.netlify.app/

This weeks project gave us a great start of working with react. We were able to meet up in pair programming sessions every day and both appreciated the great team work very much. We started by creating a component for the album and then also decided to make a component for the header, and mounted it in the app, where we had to mappings. Last minute, we also added the commas and &'s for the artists along with the second mapping. 

We have used both grid and flexbox for this project and both agree that we find grid easier and flexbox a bit more challenging. 

We'd both appreciate even more practice of the best use of flexbox and also more practice on mapping and iteration, but we feel this weeks work has run smoothly all along.


code/src/Album.js
```diff
@@ -0,0 +1,24 @@
...
+            <div className="icons">
+            <img className="heart-icon" src="icons/heart.svg" alt="" />
+            <img className="play-icon"  src="icons/play.svg" alt="" />
+            <img className="dots-icon"  src="icons/dots.svg" alt="" />
+            </div>
+        </div>
+            <h2 className="albums-name">{props.name}</h2>
+            <div>
+            <a className="link" href={props.externalurls}>
+            <h3 className="artist-name">{props.artist}</h3></a>
```
KarolCislowski, line 19: If you have more then 1 artist, they should be displayed in a row. 
image

code/src/App.js
```diff
@@ -1,12 +1,28 @@
 import React from 'react'
 import data from './data.json'
+import {Album} from './Album.js'
+import {Header} from './Header'
+import './index.css'
+
 
 console.log(data)
```
KarolCislowski, line 8: remember to remove console.logs  from your code after you finish debugging your code.

code/src/Header.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+
+export const Header = (props) => {
```
KarolCislowski, line 3: You do not need to declare the props if you do not use them.

code/src/Album.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+
+
+export const Album = (props) => {
+    return (
+        <div className="card"> 
+          <div className="picture">
```
KarolCislowski, line 7: Your image should be still darkened when you hover over any of the icons. (you can experiment with opacities of backgrounds of every layer)

KarolCislowski (changes requested):
You are very close to reaching all goals of this assignment. Please fix hover effects and artist display issues 😸

KarolCislowski (approved):
😸

PR #116, opened 2021-03-14
Project music releases - Sofia & Felicia 

Sofarz:
Link to our deployed project: https://project-music-releases2021.netlify.app

Project by Felicia Liabäck Löwstett and Sofia Arzt Wallén


code/src/App.js
```diff
@@ -1,12 +1,19 @@
 import React from 'react'
 import data from './data.json'
+import Header from './components/Header'
+import AlbumCard from './components/AlbumCard'
```

code/src/components/Artists.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+
+const Artists = (props) => {
```
KarolCislowski, line 3: destructure your props like this:

```jsx
}
```

or like this:

```jsx
  //
}
```

KarolCislowski (approved):

PR #117, opened 2021-03-14
Project Music Releases - Lars Söderman & Anna Lindgren

magicvilla:
Hello! 

https://elegant-hawking-15cb3f.netlify.app/


code/src/Album.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import { Artist } from './Artist'
+
+
+export const Album = (props) => {
+    if (props.albumType === "album") {
+        console.log(props)
```
KarolCislowski, line 7: remember to remove console.logs  from your code after you finish debugging your code.

code/src/Album.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import { Artist } from './Artist'
```
KarolCislowski, line 2: There is a convention that says that we should divide outer dependencies (those located in node_modules) from local dependencies (all local files that we created - components, functions, data such as arrays) by one empty line. Moreover, style.css should be separated as well and located below local dependencies.

code/src/App.js
```diff
@@ -1,12 +1,51 @@
 import React from 'react'
 import data from './data.json'
+import { Album } from './Album'
+import { Heading } from './Heading'
+import { Singles } from './Singles'
 
 console.log(data)
```
KarolCislowski, line 7: again logs. You really always need to remove it from the finished version of your code

KarolCislowski (approved):

Great Job!

PR #118, opened 2021-03-14
React Project Anna-Kaisa & Noemi

NoemiSalome:
Project Anna-Kaisa / Noemi Frischknecht

Link to our deployed project:
[https://project-musicrelease-anna-kaisa-noemi.netlify.app](url)

Reflections on how this week's project turned out:
The first struggles we stumbled over was getting the data displayed on the browser. Working with the ressources we got from wednesdays lecture got us on the right way but it was the exchaning with a teammember, that brought us there in the end even though we went for a different approach than our teammember in the end. After that we had a rough time with the styling (flexbox), even though we both had quite a good understanding of the flexbox model. Especially getting several artist seperated with "," and "&" in line took quite some time. 

Even though there were several different ways to solve this weeks project we could find answers on stackoverflow and from our teammembers. Also the demo on friday helped us getting faster further with our project. 

We were having a harder time with flexbox than expected. Some of our teammates went with grid and they felt like it was easier. Now we are wondering when to go with grid and when with flexbox. Are there cases when one is better than the other?

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -15,9 +15,8 @@ export const App = () => {
     <>
       <div className="main">
         <Header />
-        
         {albumArray.map(item => (
-        <Album key={item.id} albumContent={item} />
+          <Album key={item.id} albumContent={item} />
         ))}
       </div>
     </>
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -6,24 +6,24 @@ import Header from './components/Header';
 
 import data from './data.json'
 
+console.log(data);
 
-
+const albumArray = data.albums.items; 
 
 export const App = () => {
-  console.log(data);
-
   return (
     <>
-    <Header />
-    <Album />
+      <Header />
+      <div className="whole-container">
+         {albumArray.map((album) => {
+            return (
+            <Album key={album.name} name={album.name} artist={album.artists[0].name} picture={album.images[0].url} externalurls={album.artists[0].external_urls.spotify} albumurl={album.external_urls.spotify} />
+          )
+          })}
+      </div>
     </>
-    
-   /* <div>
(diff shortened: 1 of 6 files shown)


code/src/App.js
```diff
@@ -1,12 +1,25 @@
...
-      Find me in src/app.js!
-    </div>
+    <>
+      <div className="main">
+        <Header />
+        {albumArray.map(item => (
+          <Album key={item.id} albumContent={item} />
+        ))}
+      </div>
+    </>
```
KarolCislowski, line 22: If your component returns only one root element, there is no need to wrap it around React Fragment.

code/src/components/Album.js
```diff
@@ -0,0 +1,55 @@
...
+                                                <p className="comma"> {index > 0 && index !== props.albumContent.artists.length - 1 && ' , '}</p>
+                                                <p className="comma"> {index !== 0 && index === props.albumContent.artists.length - 1 && ' & '}</p>
+                                                <h2 key={artist} className="artists">{artist.name}</h2>
+                                                </div>
+                                            </React.Fragment>
+                                        )
+                                        })}
+                                    </a>
+                                </div>
+                        </div>
```
KarolCislowski, line 46: The text under your album cover needs to be aligned to the left to meet layout requirements. 🎨 

You ware also instracted to use fallowing fonts:
• Album title - 14px Helvetica #ffffff
• Artist name - 14px Helvetica #a0a0a0

KarolCislowski (changes requested):
Please adjust your styling to meet layout requirements 😸

KarolCislowski (approved):
Great 😸

PR #121, opened 2021-03-14
Project Music releases by Anna Lenksjö & Yennie Leek

annalenksjo:
Link to my deployed project:
https://quizzical-noether-570a46.netlify.app/ 

It was fun to do something new, and we think the structure of React makes sense. It was difficult in the beginning of the week but in the end we are happy with our final product. It will be fun to do more React!

More about how Javascript and React matches together, more of a comparison, where are the overlaps. If there is something specific in Javascript that is mostly used in React.


code/src/App.js
```diff
@@ -1,12 +1,56 @@
 import React from 'react'
 import data from './data.json'
+import { Albums } from './Components/Albums'
+import { Heading } from './Components/Heading'
+import { Artists } from './Components/Artists'
+import { AlbumTitle } from './Components/AlbumTitle'
+import { TypeOf } from './Components/TypeOf'
+import { Footer } from './Components/Footer'
+import './App.css'
```
KarolCislowski, line 9: There is a convention that says that we should divide outer dependencies (those located in node_modules) from local dependencies (all local files that we created - components, functions, data such as arrays) by one empty line. Moreover, style.css should be separated as well and located below local dependencies.

KarolCislowski (approved):

Great Job!

PR #123, opened 2021-03-15
Project music release

dannuzak:
Link to my deployed project:

https://music-release-site-by-maria-and-daniela.netlify.app

It was exciting to start learning how to work with React. It was challenging to understand on my own how to fetch data from json file. It was also challenging to understand how to make work with the filter on hover in combination with scaling the play button. We received help from classmates and that was hugely appreciated.

 When we started the project on Tuesday we felt stuck already from the beginning as we couldnt figure out on our own how to fetch data with map method from json and inject it in a React component. I had wished this information already from Monday  and explained in more detailed. I would also have wished to talk about CSS and how to think when two hover effects affect each other when layered in the same elements.


code/src/App.js
```diff
@@ -1,12 +1,38 @@
 import React from 'react'
 import data from './data.json'
+import Header from './components/Header'
+import Album from './components/Album'
```

code/src/components/Album.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
+
+export const Album = (props) => {
```
KarolCislowski, line 3: destructure your props like this:

```jsx
}
```

or like this:

```jsx
  //
}
```

KarolCislowski (approved):

Great Job!

PR #144, opened 2021-10-15
annadimi-ebbadelsol-project-music-releases

anndimi:
Link to my deployed project:


code/src/index.css
```diff
@@ -1,13 +1,183 @@
...
+	display: block;
+	margin-bottom: 10px;
+}
+
+.album-cover-overlay {
+	opacity: 0;
+	position: absolute;
+	top: 50%;
+	left: 50%;
+	transform: translate(-50%, -50%);
```
jakobxlindstrom, line 58: Because your position absolute is absolute to the album-image-container. Your overlay is already on top before you add line 56-58. You dont need these lines.

code/src/index.css
```diff
@@ -1,13 +1,183 @@
...
+	padding: 112.5px 42.5px;
```
jakobxlindstrom, line 59: If we understand it correctly, you have added this very specific padding to center the icons on the album. Instead you could have used the flex attributes to align and center it on the card.

code/src/index.css
```diff
@@ -1,13 +1,183 @@
...
+@media (min-width: 1903px) {
+	header {
+		width: 1770px;
+	}
+}
+
+@media (min-width: 1940px) {
+	header {
+		width: 1900px;
+	}
```
jakobxlindstrom, line 182: For the future instead of adding this amount of media queries you could have benefited by using a percentage 
 on margin-left or the entire width of the header for the header text. It would then follow seamlessly and stay inline with the top left album.

PR #148, opened 2021-10-15
Week 9: Music App by Pinar Demirel & Daniel Vestin

dandeloid:
Link to my deployed project:


code/src/App.js
```diff
@@ -1,12 +1,30 @@
...
-      Find me in src/app.js!
+    <div className="site-container">
+      <Header />
+        <div className="container">
+          {data.albums.items.map((item) => {
+            return <Album
+              img = {item.images[1].url}
+              title = {item.name}
+              artist={item.artists.map((item) => 
+                <a className="artistLink" href={item.external_urls.spotify} key={item.name} target="_blank" rel="noopener noreferrer">
```
efstasia, line 18: One suggestion is to use id instead of name as a key in the future since it can cause confliction if the same artist has more than one release. Like this  `key = {item.id}` .

code/src/components/Album.js
```diff
@@ -0,0 +1,20 @@
...
+            <img className="dots-icon"  src="icons/dots.svg" alt="dots icon" />
+          </div>
+        </div> 
+          <a className="albumLink" href={props.albumLink} target="_blank" rel="noopener noreferrer">
+            <h2 className="title">{props.title}</h2>
+          </a>
+          <h3 className="artist">{props.artist}</h3>
+      </article>
+      )
+}
```
efstasia, line 20: Looks nice and clean, easy to understand!

code/src/index.css
```diff
@@ -11,3 +14,124 @@ code {
...
+  color: grey;
+  border-top: 1px solid rgb(60, 60, 60);
+  text-align: center;
+  padding-top: 16px;
+}
+
+/* Media queries */
+@media (min-width: 768px) {
+  .container {
+    display: grid;
```
efstasia, line 118: No need to display grid again here since you have it in mobile view.

code/src/index.css
```diff
@@ -11,3 +14,124 @@ code {
...
+    justify-items: center;
+    grid-template-columns: repeat(2, 1fr);
+    grid-template-rows: repeat(2, 1fr);
```
efstasia, line 121: Is it necessary to have grid-template-rows when you already have grid-template-columns?

efstasia (commented):
The page looks and works like it should! We like that you added an extra media query for smaller desktops. Good job on getting the red requirement in! The page is responsive and the code looks nice and clean. 
Good job on managing to keep the code to a minimum and still getting the same result as we did, really impressive! ⭐ 

// Katie and Sofia

PR #150, opened 2021-10-15
Sarah Mottram & Johanna Mannestål Music Releases

JohannaMJ:
Pair programming assignment to build a latest music releases website with React and JSX, using multiple components, made by Sarah Mottram and Johanna Mannestål.

We started by creating the structure for our first component, and then based on this structure we decided what data we would need to give that component when we mounted it. We then continued in this same pattern for the next component. We used stack overflow a lot during the process.
If we had more time we would look into how to do the live fetch in React so that the music releases would be up to date.

https://tiger-music-releases.netlify.app/


code/src/style.css
```diff
@@ -0,0 +1,151 @@
...
+}
+
+.artist:hover {
+  text-decoration: underline;
+}
+
+.icons-overlay {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-evenly;
```
Asivol93, line 75: I don't think you need theese lines of code when I check the dev tools. They don't seem to do anything since you already saying that it should be position: absolute; top: 50%; left:50%; which centers it :)
sarah Johanna

code/src/style.css
```diff
@@ -0,0 +1,151 @@
...
+  width: 60px;
+  height: 60px;
+  transition: all 0.3s ease-in-out;
+}
+
+.play:hover {
+  opacity: 1;
+  transform: scale(1.2);
+  cursor: pointer;
+}
```
Asivol93, line 118: Turned out real clean!

code/src/components/Artists.js
```diff
@@ -0,0 +1,33 @@
...
+        return (
+          <a
+            key={artist.id}
+            href={artist.external_urls.spotify}
+            className='artist'
+          >
+            {artist.name}
+            {getSeparator(i, artistsArray.length)}
+          </a>
+        );
```
Asivol93, line 27: Nice that you did it with javascript, the css solution I went for was a bit confusing. I know what this code means but maybe could have a comment for explanation for others to know if they ever look at the code in the future :)

Asivol93 (commented):
You did a great job with the first React project and it looks just as clean as the project brief! Also passed a validation tool and works fine in chrome, firefox and edge! :)

I left some comments in the code but nog bugs or anything which is great!

## **🔵  Blue Level (Minimum Requirements)**

- Your app has have components
- *Your page looks as close as possible to the design provided*. (except the font)
- You're displaying the following for each album:
    - album cover image
    - album name
    - the name of each artist involved with a comma between
- Hover effects:
    - When you hover over an album, it goes darker, and you're showing a play button, a favourite button, and the ellipsis
    - When you hover over the play button, it grows 
    - Hovering an artist name or an album name matches the design sketch
- When you click on the album, it links to the album's external url in the data
- When you click on an artist, it links to the artist's external url in the data
- Your page is responsive:
    - showing 4 albums per row on desktop
    - 2 per row on tablet
    - 1 per row on mobile.

code/src/components/Album.js
```diff
@@ -0,0 +1,22 @@
...
+
+const Album = (props) => {
+  const { src, title, artistsArray, albumURL } = props;
+  return (
+    <article className='song-card'>
+      <div className='album-art-container'>
+        <img className='song-image' src={src} alt='album cover' />
+        <IconsOverlay />
+      </div>
+      <a href={albumURL}>
```
Asivol93, line 14: Could be nice to add target="blank" to get the song in a new tab and not leave the site. Also you could add rel="noreferrer noopener" for security reasons (just learned from Lousanne!). Read more here: https://www.reliablesoft.net/noreferrer-noopener/

PR #158, opened 2021-10-15
Music releases by Maria Petersson & Anna Thunberg

annathunberg:
Link to my deployed project:

https://banana-music.netlify.app/

We are proud of our project. There are some small adjustments we could have made, but are happy with the turnout.

We are intrigued by the fact that our artist and album links are in the opposite order, even though they were mounted in app.js according to the design.


code/src/App.js
```diff
@@ -1,12 +1,36 @@
...
+import React from "react";
+import "index.css";
+import data from "./data.json";
+import Album from "./components/Album";
+import Artist from "./components/Artist";
+import Header from "components/Header";
 
 export const App = () => {
+  const musicData = data.albums.items;
+  console.log(musicData);
```
waliem, line 10: console.log survivor!

code/src/TODO
```diff
@@ -0,0 +1,39 @@
+Each album:
+    - album cover image ✅
+    - album name ✅
+    - the name of each artist involved with a comma between ✅
+    fix the widht of the artist!✅
+
+Hover effects:
```
waliem, line 7: is this supposed to be here? Since it's a guide for your work process, maybe it's not necessary to keep it here when the project is done?

code/src/components/Album.js
```diff
@@ -0,0 +1,38 @@
+import React from "react";
+
+const Album = (props) => {
+  // console.log('PROPS', props);
```
waliem, line 4: another survivor!!!

PR #170, opened 2021-10-17
Aschwin & Christoffer's React Music Releases

vrill:
Link to our deployed project:
https://brave-einstein-108ec4.netlify.app/

Good challenge and it turned out pretty sweet, see the readme for more detail. :)

The import/export situation is quite strict, and the `key` is still something we feel needs to be understood better.


code/src/index.js
```diff
@@ -1,6 +1,10 @@
...
-import './index.css'
+
+import { Header } from './components/Header'
 import { App } from './App'
 
-ReactDOM.render(<App />, document.getElementById('root'))
+import './index.css'
+
+ReactDOM.render(<Header />, document.getElementById('header'));
+ReactDOM.render(<App />, document.getElementById('root'));
```
Jsfrulle, line 10: Hi Guys, 

Too start we will say GOOD WORK! The features is working and the links leads you to right artist/album! 
you need to take a closer look on you styling - the hr(line), album and header,  most line up under each other! The display of the album is giving a gap on last albums so its seams like an album is missing(its not but looks like it) This we think you can solve easy with flex-justify: flex-start! Ande the last thing is that you need to adjust the margin(gose hand in hand with the line up of the elements)! Look at the images technigo gave us a last time!

- [YES ]  Am I able to understand the code easily?
- [NO ]  Does the code have lots of duplicates? Could it be tidied up?
- [WORKS ]  Does it work or does anything look broken?
- [Just fix little styling things and you are there! ]  Does it follow the general and blue requirements?

PR #182, opened 2022-03-10
Week 9: Jessica Sandler, Laura Sjölander and Nadia Lefebvre

nadialefebvre:
Link to our deployed project:
Project deployed here: https://music-releases-spotify.netlify.app/

Our reflections on how this week's project turned out:
This week's project was to practice with React and components while using data from Spotify. We had to build a page displaying new music releases. Project done by Jessica Sandler, Laura Sjölander and Nadia Lefebvre.

The project had to be done with a pair-programming approach. In order to fully learn how to do it, we agreed to try creating components and displaying some result in a browser on our own before sharing what we had done. After we were all three able to do that, we coded along the final version. We decided to split everything in small components since this structure makes the code simpler to read and more reusable. In fact, when we added the filter to split albums & singles in two sections, there were only some lines to add in our main file since our components were fully reusable.


code/src/index.css
```diff
@@ -1,13 +1,178 @@
...
+  column-gap: 20px;
+  row-gap: 40px;
+}
+
+.card {
+  width: 300px;
+}
+
+.image-box {
+  width: fit-content;
```
rijad90, line 48: nice job making the image fit the box, inspirational!

PR #188, opened 2022-03-11
Music app by Dorothea and Kristiina

kolkri:
https://music-app-dorothea-kristiina.netlify.app/


code/src/App.js
```diff
@@ -1,12 +1,54 @@
...
+          return <Albums key={singleAlbum.id} articleDetails={singleAlbum}/>
+          else return null
+        })}
+      </section>
+
+      <h2>New Singles</h2>
+
+      <section>
+        {data.albums.items.map((singleSingle) => {
+          if (singleSingle.album_type === "single")
```
MarianneArdin, line 34: There is a different way or doing this: instead of using "else return null", try -catch method.

code/src/index.css
```diff
@@ -1,13 +1,222 @@
...
   -moz-osx-font-smoothing: grayscale;
+  cursor: default;
+  font-family: "Helvetica", Sans-Serif;
+
+}
+
+section {
+  display: flex;
+  flex-wrap: nowrap;
+  overflow-x: scroll;
```
MarianneArdin, line 23: We really like the carousel feature!

code/src/index.css
```diff
@@ -1,13 +1,222 @@
...
+.artist-container a:nth-last-of-type(2)::after {
+  content: " & ";
+  color: #A0A0A0;
+}
+
+.album-cover {
+  position: relative;
+  width: 100%;
+  transition-duration: 1s;
+  /* display: flex; */
```
MarianneArdin, line 122: Line 122 appears to be commented out. If its not used it can be removed.

code/src/components/Playlists.js
```diff
@@ -0,0 +1,32 @@
...
+export const Playlists = (props) => {
+  
+    return (
+    <article className="album">
+      {/* ===== ALBUM COVER ===== */}
+      <div className="album-cover">
+
+      
+        <a href={props.articleDetails.external_urls.spotify} target="_blank" rel="noopener noreferrer">
+          <img className="album-image" src={props.articleDetails.images[0].url} alt="album cover" />
```
MarianneArdin, line 12: The alt needs to have a clear name for the screen reader, for example "album name" instead of "album cover" by fetching from data.json.

PR #189, opened 2022-03-11
Suki Phan and Marianne Ardin

MarianneArdin:
Link to my deployed project: https://suki-marianne-music-app.netlify.app/


code/src/index.css
```diff
@@ -5,9 +5,107 @@ body {
...
+
+
+.container {
+  gap: 10rem;
+  margin: 4rem 2rem;
+}
+
+.displayAlbum {
+  display: grid;
+  grid-template-columns: repeat(auto-fit, minmax(35rem, 1fr));
```
tiiliu, line 33: We noticed that in the desktop view it is showing only three images when in the instructions they are requiring 4 images for desktop. The mobile and tablet view works but we think you could add a media query for the grid-template-columns in desktop view (starting for example from 1024px). Right now your site breaks into 3 images in 1200px and into 4 images at 1500px which is very big. :)

code/src/components/AlbumName.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+
+const AlbumName=(props) => {
+    return(<a className='album-name' href={props.name.external_urls.spotify}>{props.name.name}</a>)
```
tiiliu, line 4: We noticed that the link in play-icon was opening in a new browser tab, but this one not. Maybe it could be better to have a consistent way to do it with the links, either opening all in a new tab or all in the same tab? Just as a thought :)

tiiliu (changes requested):
Suki and Marianne, you've created a beautiful looking site! The images and titles are very nicely lined, the grid-template-columns work well. Your code was very clean and well-written. We noticed some blue-level requirements which were not filled and we added comments about them inside your code. We hope you find them helpful :) Great job! 🥇 🐱 
Code review done by Mimmi Fordal Uddin and Tiina Liukkonen 🌸

code/src/index.css
```diff
@@ -5,9 +5,107 @@ body {
     sans-serif;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
+  background-color: black;
+}
+
+html {
+  font-size: 62.5%;
```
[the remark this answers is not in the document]
MarianneArdin: The font-size refers to how the rem in CSS 1 rem=10px.

PR #198, opened 2022-03-11
Music Releases - Michael

michaelchangdk:
Link to my deployed project:
https://project-spotify-week9.netlify.app/
https://technigo--project-spotify-week9.netlify.app/

I'm really proud of my components and how elegant they are (in my opinion at least) - especially on the carousel / flickity version (the primary one). Think I could have spent some more time in the beginning learning instead of creating the project as quickly as possible, and then going back and re-writing multiple components several times. But I think it was a valuable learning experience creating so many different versions of components and trying to fit them together.

Not much. I feel like it's pretty straight forward.


code/src/components/legacy/Albums.jsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import data from '../data.json'
+import heart from '../icons/heart_white.svg'
+import play from '../icons/play_white.svg'
+import dots from '../icons/dots_white.svg'
+
+let albums = data.albums.items.filter((album) => album.album_type === "album")
+// console.log(albums)
```
NabeelMansour, line 8: maybe it would be a good idea to remove the console.log, before end product. Just a thought

code/src/components/legacy/Playlist.jsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import stretchGoal from '../stretch-goal.json'
+import heart from '../icons/heart.svg'
+import play from '../icons/play.svg'
+import dots from '../icons/dots.svg'
+
+let playlists = stretchGoal.playlists.items
+// console.log(playlists)
```
NabeelMansour, line 8: Same goes here with the log

PR #201, opened 2022-03-13
Sofia Ringstedt & Jacob Englund New Releases project

sofiaringstedt:
https://sofia-jacob-project-music-releases.netlify.app/

### DIFF
--- a/code/public/index.html
+++ b/code/public/index.html
@@ -13,7 +13,7 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <title>Technigo React App</title>
+    <title>Music releases</title>
   </head>
 
   <body>
--- a/code/public/index.html
+++ b/code/public/index.html
@@ -13,7 +13,7 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <title>Music releases</title>
+    <title>New Releases</title>
   </head>
 
   <body>
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -3,16 +3,14 @@ import MusicInfo from 'components/MusicInfo'
 import data from 'data.json'
 import Header from 'components/Header'
 
-//console.log(data)
+/* Rendering the main structure of website */
 
 export const App = () => {
 	return (
-    <>
-    <div class="wrapper">
-        <header><Header /></header>
-    <div className="page-wrapper">
-      <MusicInfo musicList={data} />
+    <div className="wrapper">
+      <header><Header /></header>
+      <div className="page-wrapper">
+        <MusicInfo musicList={data} />
+      </div>
     </div>
-    </div>
-    </>
 	)}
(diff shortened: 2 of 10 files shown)


code/src/App.js
```diff
@@ -1,12 +1,17 @@
...
+
+/* Rendering the main structure of website */
 
 export const App = () => {
-  return (
-    <div>
-      Find me in src/app.js!
+	return (
+    <div className="wrapper">
+      <header><Header /></header>
```
jenniferwvng, line 12: Nice use of a Header component! Maybe you could also put the header html tags in the component?

code/src/components/MusicInfo.js
```diff
@@ -0,0 +1,18 @@
...
+
+import { Albums } from './Albums'
+
+/* Creating the main album wrapper and mapping through it, to be accessed in Albums.js */
+const MusicInfo = (props) => {
+  return (
+    <section className="albums-wrapper">
+    {props.musicList.albums.items.map((item) => {
+         return <Albums key={item.id} albumTitle={item.name} 
+         image={item.images[1].url} artists={item.artists[0].name}
```
jenniferwvng, line 11: Nice, it's easy to miss out on but some albums actually have more than one artist (e.g. artists[0], artists[1] etc.). One idea is that you could make another map for the artists array, and perhaps separate the artist info into another component to import to make it easier to follow the mapping.

code/src/index.css
```diff
@@ -1,13 +1,133 @@
...
+/*ICON STYLING*/
+.iconsWrapper {
+  display: none;
+}
+  
+.iconsWrapper .icons {
+  display: inline;
+}
+  
+.iconsWrapper {
```
jenniferwvng, line 96: Nice! Clear

code/public/index.html
```diff
@@ -13,7 +13,7 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <title>Technigo React App</title>
+    <title>New Releases</title>
```
jenniferwvng, line 16: Nice that you remembered to change this!

code/src/components/Hover.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+const Hover = (props) => {
```
jenniferwvng, line 3: Maybe just a matter of taste but personally think it would be easier to follow if naming it Icons or IconHover

jenniferwvng: Also seems like the props is not used in this component :)

jenniferwvng (commented):
Beautifully styled page with responsiveness, really like the icon hover effect. URLs are working as well. One thing to note is that the code only displays one artist at the moment, even if there are more than one per album. One solution could be to map over the artists array inside the already exisiting map. Also to add comma signs after each artist, which could be done inline when displaying it e.g. {artistname + ", "}.  Or, check out how the built in index in .map() could help with the conditional of not displaying a comma after the last artist name.

Great work!

code/src/components/MusicInfo.js
sofiaringstedt: ! Totally did not see this part - thank you so much for pointing this out! On it right away!

PR #225, opened 2022-09-23
Project Music Releases | Annika Gadman, Charlotte Johansson & Amanda Wellington

a-well:
Link to our deployed project: [projectmusicreleases.netlify.app](https://projectmusicreleases.netlify.app/)


code/src/components/Icons.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+export const Icons = () => {
+  return (
+    <div className="icons-container">
+      <img className="heart-icon icons" src="../icons/heart.svg" alt="heart-icon" />
+      <img className="play-icon icons" src="../icons/play.svg" alt="play-icon" />
```
petrasson, line 7: Perhaps add a href here to spotify?

code/src/components/AlbumCard.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { Icons } from './Icons';
```
petrasson, line 2: Out of curiosity, why do you add { } here? Is it bc you import many icons?

code/src/components/AlbumCard.js
```diff
@@ -0,0 +1,30 @@
...
+import ArtistName from './ArtistName';
+
+const AlbumCard = (props) => {
```
petrasson, line 5: A reflection; what is the best practice and why? When shall we pass on (props) to a component and write props.name and when shall we pass on all the names instead const AlbumCard = ({img, URL, name})?

code/src/components/AlbumCard.js
```diff
@@ -0,0 +1,30 @@
...
+      <p className="album-name">
+        <a href={props.url}>{props.name}</a>
+      </p>
+      {props.artists.map((artist, i) => (
+        <>
+          <ArtistName
+            key={artist.id}
+            name={artist.name}
+            url={artist.external_urls.spotify} />
+          {i + 1 < props.artists.length && <span className="artist-name" style={{ paddingLeft: 3, paddingRight: 3 }}> , </span>}
```
petrasson, line 23: We found a solution (https://stackoverflow.com/c/technigo/questions/2239) to add , and & between Artists

This how we solved it:  
https://github.com/petrasson/project-music-releases/blob/master/code/src/components/Artists.js

petrasson (approved):
Great job!! 🙌 Your code looks clear and easy to follow.
Nice structure and choices of the components. 
We have added some questions and a solution on the ", & between artists" if you want to add that.

code/src/components/AlbumCard.js
a-well: Best practice is to deconstruct the props like you did there in the last example, since it makes the code much easier to follow. Will update code now 🙏🏻

code/src/components/AlbumCard.js
a-well: No, the number of icons does not affect how the import should be done. 
The reason why there are curly brackets there is because the component is being imported as a named component, and for named components you need to add `{}` around the name. It should actually be a default import/export though since there is only one thing to export in Icons.js, so will fix that. Good spot, thank you!

code/src/components/AlbumCard.js
a-well: Wonderful, thank you! 😀

PR #226, opened 2022-09-23
Therese Ånmark & Tina Bruce - project-music-releases

brucette:
Link to my deployed project:

https://soft-palmier-811892.netlify.app/


fannycardell (commented):
Hello ladies!
The application works fine. Nice job with the ", &" between the artists and the transition on the hovering of the albums and the links works as it should!

Small thing with the CSS: 
The icons only appears when hovering the center of the album image, and they are slightly to the right side. 
In the desktop view the last two albums are centered to the screen instead of staying fixed to the left.
We have commented this more in detail in the code=)

Overall great job with a functional application and well written code!

PR #227, opened 2022-09-23
Project Music Release - Viktor & Kartika

katirum:
Link to my deployed project:
https://v-k-music.netlify.app/


code/src/App.js
```diff
@@ -1,12 +1,34 @@
...
+          <Cover key={item.id} data={item} img={item.images}/> 
+          <Albumname Albumname={item.name} urlAlbumname={item.name}/>
+          <Artistname artists={item.artists} artistUrl={item.artists}/>
+        </div>
+        </>
+      })}
+      </div>
+      <div className='Footer'>
+     <h3>Project Music Release for Technigo by Viktor & Kartika</h3>
+     </div>
```
MariaWellander, line 29: Nice idea with a footer, and it looks good! Could consider not making an extra div around the h3, and just put the "footer" class directly on h3 as well.

code/src/App.js
```diff
@@ -1,12 +1,34 @@
...
     </div>
   );
 }
+
+//we're declaring a variable "data" and {item} is the value. so we connect the varibale Line 15
```
MariaWellander, line 34: If you want to add comments in JSX you have to wrap the comment inside {}.
Adding the comment at the end is a bit problematic since the line number might change by the time you are finished with the project, or if you come back to the project to make changes you will have to keep in mind to also change your comments. :)

code/src/components/Header.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+
+const Header = () => {
+    return (
+        <a href='https://open.spotify.com/genre/new-releases-page' rel="noopener noreferrer"><h1>New albums & singles</h1></a>
+)}
+
+export default Header;
```
MariaWellander, line 8: This component makes a link out of the entire div that wraps the header, which feels a bit risky when navigating the page that you will accidentally click this big link. If you wanted to only have the actual text be a link you could make the h1 into display: inline, then move the padding and border styling to the 'Header' div.

code/src/index.css
```diff
@@ -5,9 +5,169 @@ body {
...
+
+/* For the whole music albums */
+.music-container {
+  display: grid;
+  grid-template-columns: 100%;
+  margin: auto 0;
+  padding: 2%;
+}
+
+.theAlbums {
```
MariaWellander, line 25: The casing convention for CSS class names is typically kebab-case. Ultimately it doesn't matter which casing you go for, but probably best to be consistent throughout a project especially if other people will work on the same project. :)

code/src/index.css
```diff
@@ -5,9 +5,169 @@ body {
...
+  margin: 20px auto;
```
MariaWellander, line 26: Nice job with centering the albums. This margin: 20px auto; works magic to center the albums on mobile, I will steal this for our project 🙂

code/src/App.js
```diff
@@ -1,12 +1,34 @@
...
+    <div className='all-container'>
+      <div className="Header"> 
+       <Header /> 
+      </div>
+      <div className='music-container'>
+      {data.albums.items.map (item => {
+       return <>
+        <div className='theAlbums'>
+          <Cover key={item.id} data={item} img={item.images}/> 
+          <Albumname Albumname={item.name} urlAlbumname={item.name}/>
```
MariaWellander, line 21: When we are clicking the album title the link is broken :( It seems like the urlAlbumname variable gets the wrong data. Try this instead:
`item.external_urls.spotify`

MariaWellander (commented):
• Am I able to understand the code easily?
Yes! Everything is very easy to understand :D Though, we're really interested in hearing about how you found the solution for the condition in ArtistName - wasn't that easy to get a grip on but it works perfectly!

• Does the code have lots of duplicates? Could it be tidied up?
Good job! No duplicates :)

• Does it work or does anything look broken?

Everything works, except from a small bug with the albumname link.

• Does it follow the general and blue requirements?

It does follow all the requirements, except from a small error with not having 14px for the album title nor using the Helvetica font for the artist name and album title.

• To sum it up

Nice that you managed to match the design by getting the line underneath h1 easy to miss that one!
Also nice with the extra effect: when you hover over the heart and dots buttons they are growing.
Great job with the comments too - we like them a lot! Be super proud of yourselves for this work!!

PR #229, opened 2022-09-24
Jessica and Cecilia music release site

jesshansson:
Week 5 project music releases

https://top50music.netlify.app/

Jessica Hansson & Cecilia Rådén


code/src/index.css
```diff
@@ -5,9 +5,180 @@ body {
...
+  display: flex;
+  justify-content: space-around;
+  align-items: center;
+  top: 50%;
+  left: 82%;
+  display: none;
+  transform: translate(-50%,-50%);
+  z-index: 5;
+  width: 90%;
+}
```
jessikalind, line 88: Like you mentioned on our Monday meeting, there are some issued with the display of the icons. They are not centered, and when you use a very large screen they move outside of the album cover img. We're not really sure exactly what's wrong, but we find it peculiar that the left percentage is so high (our solution is "left 4%"). Also, you probably should not use two "display" in the same {}. This probably explains why your "justify-content" and "align-items" do not really work.

code/src/index.css
```diff
@@ -5,9 +5,180 @@ body {
...
     monospace;
 }
+
+h1 {
+  color:white;
+  margin-left: 5%;
+  margin-right: 5%;
+  border-bottom: 1px solid #65686e;
+  font-size: 27px;
+}
```
jessikalind, line 23: We noticed that the h1 is not aligned with the body (the display of all the albums) as it is in the design. Perhaps this could be solved with the use of an outer and inner container for the entire site 🥳

code/src/components/Album.js
```diff
@@ -0,0 +1,43 @@
...
+          </a>
+        })}
+      </div>
+    </div>
+  );
+};
+
+
+/* We do a .map on artists to be able to get all the artists names who
+are involved, since the names are inside an array of objects */
```
jessikalind, line 41: Very nice explanation! ✨

code/src/skiss.js
```diff
@@ -0,0 +1,46 @@
...
+            </div>;
+        })}
+      
+        </div>
+    
+    </>
+  );
+}; 
+}
+*/
```
jessikalind, line 46: Perhaps this file could be removed :)

jessikalind (commented):
Great job on the week 5 assignment! It is a great looking page, that in most instances looks like the given design 🥳 Great work on the use of the map-function and getting the connection to the json to work!

A general remark is however that there seems to be some issues with the layout in some screen sizes. In mobile view (we had the exact same issue) the albums are aligned to the left instead of centered, and when going from tablet-view to desktop-view, the last column to the right get a bit cropped. On a large screen, the album images spread out all over the page. Maybe adding a max-width and an inner-wrapper could help with this ☺️

A last thought is that the Album-component perhaps could be split into several components (for example, one for the artists' names and one for the album names). It however works fine as it is now, but adding more components could make the code more clean. 

/Jessika and Linda

jessikalind (approved):
Great job on the week 5 assignment! It is a great looking page, that in most instances looks like the given design 🥳 Great work on the use of the map-function and getting the connection to the json to work!

A general remark is however that there seems to be some issues with the layout in some screen sizes. In mobile view (we had the exact same issue) the albums are aligned to the left instead of centered, and when going from tablet-view to desktop-view, the last column to the right get a bit cropped. On a large screen, the album images spread out all over the page. Maybe adding a max-width and an inner-wrapper could help with this ☺️

A last thought is that the Album-component perhaps could be split into several components (for example, one for the artists' names and one for the album names). It however works fine as it is now, but adding more components could make the code more clean. 

/Jessika and Linda

PR #232, opened 2022-09-25
Week 5: Ulrika Öhman & Arnau Monforte

Iracuzzi:
Link to my deployed project:

https://week5musicweb.netlify.app/


code/src/components/Album.js
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+
+const Album = (props) => {
+ const {images, artists , name} = props.items
```
jessnmo, line 4: great to see deconstruction here! You could make this code even simpler like this:
const Album = ({images, artists, name}) =>{}

code/src/components/Header.js
```diff
@@ -0,0 +1,17 @@
...
+
+const Header = () => {
+
+    return (
+        <div>
+
+            <h1>
+            New albums & singles
+            </h1>
+            <hr />
```
jessnmo, line 11: this is a good one, didn't thought about this tag. Instead used a border bottom in css, which in the end caused some styling problems

jessnmo (commented):
Great work guys!

PR #235, opened 2022-09-25
Music releases | Petra Söderström & Julia Östedt

juliaostedt:
Link to my deployed project:
https://music-releases-petra-julia.netlify.app/

Very difficult week, had issues understanding how everything was connected and would had liked some guidance on how to set up the structure from the beginning. 

Basics on how the npm actually works. 
How props works.


code/src/components/Album.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+
+import Images from "./Images";
+import AlbumTitle from "./AlbumTitle";
+import Artist from "./Artist";
+
+const Album = (props) => {
```
Sailornina, line 7: What do you think about using destructuring props in React? Maybe, It's a good option to use it in this case where we need to use many properties like image, URL, name, and so on 😄

code/src/components/Artist.js
```diff
@@ -0,0 +1,26 @@
...
+          return (
+            <a
+              className="title-links"
+              href={artist.external_urls.spotify}
+              target="_blank"
+              rel="noopener noreferrer"
+            >
+              <h3 className="artist-name">
+                {artist.name}
+                {props.artistArray.length - 1 > index ? ", " : " "}
```
Sailornina, line 17: Good idea!

Sailornina (commented):
Hi Julia and Petra! 
Good Job! 🤗 

Reviewing by Antonella and Saralie ⚡

PR #237, opened 2022-09-25
Music-release-project Lisa Johansson Nina Berggren

NinaBerggren:
Link to my deployed project:


NinaBerggren:
Netlify link https://project-music-release-nina-lisa.netlify.app/

code/src/index.css
```diff
@@ -1,13 +1,139 @@
...
+  text-decoration: none;
+  /* border: 1px solid red; */
+}
+
+a:hover {
+  color: #fff;
+  text-decoration: underline;
+}
+
+.App {
```
Archkrull, line 41: To not have the content spill over, it is a good idea to make an inner wrapper and specify a max width for the content. For example, all of your content is now inside the div that is "app" so if you specify a view width the content will not go passed those boundaries.

code/src/index.css
```diff
@@ -1,13 +1,139 @@
...
+@media (min-width: 667px) {
+  .App {
+    grid-template-columns: repeat(2, 1fr);
+  }
+}
+
+/* FOR DESKTOP */
+@media (min-width: 1024px) {
+  .App {
+    grid-template-columns: repeat(4, 1fr);
```
Archkrull, line 136: If  you had problems that the columns in the grid are to big, just make the columns slightly smaller, so instead of 50%-50% maybe do 49%-49%, or four times 24%.

code/src/index.css
```diff
@@ -1,13 +1,139 @@
...
+.play:hover {
+  width: 100%;
+  transition: 0.5s ease;
+}
+
+.album-title {
+  color: #fff;
+  font-weight: 700;
+}
+
```
Archkrull, line 113: To solve the problem of the artists being on different lines you could just set the property display: inline inside of .album-artist

Archkrull (commented):
It works as it should, great that you managed to make it work in such a short time!

First thing, we think Daniels code was acidentally comitted as well and it would be great if it was removed as it is a bit confusing.

Secondly, we saw that you use pixels a lot and a good idea is to do it with percent or vw/vh instread as that works better with responsiveness.
During the demo you talked about the scroll pad and the images being to big but we don't think it is the image size that is the problem as they do get smaller. We think it is a problem of there not being an inner wrapper so the images have no border to which it is contained. If you had created a div and specified the width for all of the content.

PR #238, opened 2022-09-25
Music Releases Project with Hanna and Agnes

jacoblie:
Link to my deployed project:
https://music-releases-project-week1react.netlify.app/


karlfredrikholm:
Nice with separate css files for each component! Over all easy to follow.
Also nice to import the icons instead of writing the whole path in src, made it a lot cleaner.
End result looks very close to the original. Very well done!
/Fredrik, Tara & Kim

PR #239, opened 2022-09-25
Linda Malm & Jessika Lind - spotify top50

malmen237:
Watch the site live at:

https://spotify-top50-albums.netlify.app/


code/src/App.js
```diff
@@ -1,12 +1,27 @@
...
+                                return <Albums key={item.id} data={item}/>
+                            })}
+                    </section>
+
+                </div>
+
+            </section>
+
+        </>
+    );
```
Tanmark, line 26: Clean and clear code nice work on keeping the app code to a minimum and storing most of the information in the components.

code/src/index.css
```diff
@@ -1,13 +1,146 @@
...
+        transform: scale(2.5);
+    }
+
+    .like:hover, .more:hover {
+        transform: scale(1.3);
+    }
+
+    .cover-and-btns:hover > .btn-container {
+        opacity: 1;        
+    }
```
Tanmark, line 145: Nice bit of code that also provided some inspiration for us when we were struggling!

code/src/index.css
```diff
@@ -1,13 +1,146 @@
...
+    padding: 2%;
+    align-self: center;
+    justify-self: center;
+}
+
+.album-container {
+    display: grid;
+    grid-template-columns: auto;
+    row-gap: 5vh;
+}
```
Tanmark, line 44: Maybe add a column gap (if possible) to create a bit of space between the albums

Tanmark (commented):
Overall very nice coding, like always clean and nicely structured.
Mobile first - we noticed that there is a bit too much space on the right, at 667px it was very noticable. For tablet the layout looks good from about 800px and for the desktop, at 1025px the albums are a bit bunched together. In the desktop view the albums could also take up a bit more space (be bigger) to fit the screen.
Apart from that the styling with buttons, hover effects, underlining of links etc. looks very good and the page works as intended and is in line with the brief.
/ Thérèse & Tina

PR #248, opened 2023-03-10
Project Music Releases by Amanda Sjöström and Maja Zimnoch

majazimnoch:


code/src/App.js
```diff
@@ -1,12 +1,22 @@
...
+      <Header />
+      <div className="grid-parent">
+        {data.albums.items.map((singleAlbum) => (
+          <AlbumComponent key={singleAlbum.id} album={singleAlbum} />
+        ))}
+      </div>
+    </main>
   );
 }
+export default App;
```
smirrebinx, line 22: You probably won't need to export the default App here since you have the export const App on line 10.

code/src/Components/Albums/index.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
```
smirrebinx, line 1: You might want to consider renaming the index.js-files to match what the components include such as album.js and artis.js etcetera as it can be confusing to have several index.js-files.  You could rename the files to fit the CSS-naming.

code/src/Components/Albums/index.js
```diff
@@ -0,0 +1,31 @@
...
+import Artists from 'Components/Artists';
+import styles from './AlbumStyling.module.css';
+
```
smirrebinx, line 4: It looks like you managed to import React from react through the component folder Artists and import the folder rather than the js-file.

code/src/Components/Header/index.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import styles from './HeaderStyling.module.css';
+
+const Header = () => {
+  return (
+    <div className={styles.headerContainer}>
+      <h1 className={styles.headerTitle}>New albums & singles</h1>
```
smirrebinx, line 7: To follow best practices in HTML standards you should avoid using more than one h1 on your page. We noticed that you also have an h1 on the album names. You can read more about this here:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements

code/src/Components/Header/index.js
```diff
@@ -0,0 +1,13 @@
...
+      <hr />
```
smirrebinx, line 8: What does this closing hr do? We can't see the opening tag for this.

code/src/Components/Albums/AlbumStyling.module.css
```diff
@@ -0,0 +1,29 @@
+.album-container {
```
smirrebinx, line 1: Nicely done with the styling. Good use of different CSS-files for the different components.

code/src/Components/Artists/ArtistStyling.module.css
```diff
@@ -0,0 +1,16 @@
...
+    font-family: Helvetica;
+    color: #a0a0a0;
+    margin-top: 0;
+    font-size: 14px;
+}
+
+.artistName:hover {
+    text-decoration: underline;
+    width: fit-content;
+}
```
smirrebinx, line 11: It's hard to see in the design but it looks like the artist link should change colors to white. You could add color: white; to line 8 if you want.

code/src/Components/Header/index.js
smirrebinx: Now that we looked at the CSS file for the header we understand what the hr does :-)

code/src/index.css
```diff
@@ -1,13 +1,94 @@
...
+  justify-content: center;
+  align-items: center;
+  text-align: center;
+  opacity: 0;
+  transition: 0.4s;
+  height: 100%;
+}
+
+.icon-container:hover {
+  opacity: 1;
```
smirrebinx, line 33: This makes the icons fully white.

code/src/index.css
```diff
@@ -1,13 +1,94 @@
...
+  background-color: rgb(0 0 0 / 43%);
+}
+
+.icon {
+  width: 50px;
+  margin: 4px;
+  filter: invert(1);
```
smirrebinx, line 40: Having a 1 in the invert also makes the icon fully white. If you lower the 1 to for example 0.5 you will see the color change to grey-white. It will then go to white in your hover state on line 33 as you have written opacity 1 there.

code/src/index.css
```diff
@@ -1,13 +1,94 @@
...
+
+@media (min-width: 667px) and (max-width: 1024px) {
+  .grid-parent {
+    display: grid;
+    grid-template-columns: repeat(2, 1fr);
+    gap: 20px;
+    padding: 20px;
+  }
+
+  .iconContainer {
```
smirrebinx, line 82: We noticed that the heart-icon and dots-icon move when hovering over the play-icon. Maybe you can use this icon container to control the icon's behavior instead of using width on the play button hover state.

smirrebinx (commented):
Good job! The app looks great. We added a few comments in the different files. 

Michelle and Sandra

PR #251, opened 2023-03-10
Vio and Frida - Music Release Page

Cheroptera:
https://vio-and-fridas-music-release-project.netlify.app


code/src/App.js
```diff
@@ -1,12 +1,26 @@
...
+import Album from './components/Album'
 
 console.log(data);
 
 export const App = () => {
   return (
-    <div>
-      Find me in src/app.js!
+    <div className="main-container">
+      <Header title="New albums and singles" />
```
JosefinRobertsson, line 11: Would it have been possible to mount Header without adding the "title"-part? Since there's already an h1 in the component

code/src/components/Album.js
```diff
@@ -0,0 +1,44 @@
...
+                <img
+                  className="play-button"
+                  src="./icons/play.svg"
+                  alt="play button" />
+                <img
+                  className="more-button"
+                  src="./icons/dots.svg"
+                  alt="more button" />
+              </div>
+              <img className="albumArt" src={props.img} alt={props.albumName} />
```
JosefinRobertsson, line 24: Nicely done using data to create an alt for the image

code/src/components/Album.js
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    <article className="card">
+      <div className="linked-album">
+        <a href={props.albumLink} target="_blank" rel="noopener noreferrer">
+          <div className="album-art">
+            <div className="image-container">
+              <div className="icon-container">
+                <img
+                  className="heart-button"
+                  src="./icons/heart.svg"
```
JosefinRobertsson, line 13: Smart to change the color of the icons before adding them, instead of having to change with css code

code/src/index.css
```diff
@@ -5,9 +5,173 @@ body {
...
+}
+
+.header-container {
+  grid-row-start: 1;
+  padding: 10px;
+  display: flex;
+  flex-direction: column;
+}
+
+.header-line {
```
JosefinRobertsson, line 31: Smart solution for a line (if you don't want to use an hr-tag) and it looks good!

code/src/index.css
```diff
@@ -5,9 +5,173 @@ body {
...
 
 code {
   font-family: source-code-pro, Menlo, Monaco, Consolas, "Courier New",
     monospace;
 }
+
+/* Header */
+
+.main-container {
+  display: grid;
```
JosefinRobertsson, line 20: How come you chose grid for this part? Does it make a difference? We tried removing it and see no effect

code/src/components/Album.js
```diff
@@ -0,0 +1,44 @@
...
+      </div>
+      <h3 className="artistName">
+        {props.artists.map((artist) => {
+          return (
+            <a className="artist-link" href={artist.external_urls.spotify} target="_blank" rel="noopener noreferrer">
+              <span className="artistName">{artist.name}</span>
+            </a>
+          )
+        })}
+      </h3>
```
JosefinRobertsson, line 38: Interesting choice wrapping all of this in an h3-tag instead of for example in a div. We learned something new! We had a problem with the spaces between our & and artistname, we might try it this way and see if that solves the problem without adding extra code

code/src/index.css
```diff
@@ -5,9 +5,173 @@ body {
...
+  .card {
+    width: 20%;
+  }
+}
+
+@media (min-width: 667px) and (max-width: 1024px) {
+  .card {
+    width: 45%;
+  }
+}
```
JosefinRobertsson, line 177: How come the media queries are repeated? Is it necessary?

JosefinRobertsson (commented):
Review done by [Malin Skill](https://github.com/MalinSkill) & [Josefin Robertsson](https://github.com/JosefinRobertsson)

Easy to survey your code. Good names on all the variables. We were a little bit confused about the choice to give the Header a title and use it when you mount that component, but otherwise we found your code simple to follow. Great responsiveness! Extra plus for managing to keep the media queries so short and with a minimum amount of changes needed for them to work. Great work, keep it up!

PR #257, opened 2023-03-10
Sara & Regina Team Lions

SaraBat:
https://master--heartfelt-unicorn-93a581.netlify.app/


code/src/components/AlbumList/AlbumList.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { Album } from '../Album/Album';
+
+export const AlbumList = (props) => {
+  console.log(props)
+  return (
+    <section className="AlbumList">
```
AnnikaSonnek, line 7: It seems that it's only the artist that is a link. It would be cool if you also could click the album-name and then you're brought to the album. I think you can accomplish this the same way you fetched the link for the artist. :)

code/src/components/AlbumList/AlbumList.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { Album } from '../Album/Album';
+
+export const AlbumList = (props) => {
+  console.log(props)
```
AnnikaSonnek, line 5: Good thing that you console.logged the props. Maybe you can remove the console.log when you know it works? (I don't know if they care if we still have it there.)

code/src/components/Album/Album.css
```diff
@@ -0,0 +1,58 @@
...
+.play-img:hover {
+    width: 40%;
+    opacity: 1;
+}
+
+.heart-img {
+    width: 15%;
+}
+
+.icon-img {
```
AnnikaSonnek, line 41: I see that you used the same approach as us! Nice!

code/src/components/AlbumList/AlbumList.js
```diff
@@ -0,0 +1,15 @@
...
+  return (
+    <section className="AlbumList">
+      {props.data.albums.items.map((singleAlbum) => (
```
AnnikaSonnek, line 8: Have I understood it correctly that AlbumList.js is only a component for filtering the data from data.json?

code/src/components/Artist/Artist.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import './Artist.css';
+
+const generateDelimiter = (index, length) => {
```
AnnikaSonnek, line 4: Interesting solution. Where does length come from?

code/src/components/AlbumList/AlbumList.css
AnnikaSonnek, line 1: Maybe you can remove this file since it isn't used?

code/src/components/Album/Album.css
```diff
@@ -0,0 +1,58 @@
...
+    text-align: center;
+    margin: 10%;
+    opacity: 0;
+    transition: .2s ease;
+}
+
+.imgAndIcons:hover .icon-img{
+    opacity: 0.8;
+}
+.imgAndIcons:hover .AlbumImg{
```
AnnikaSonnek, line 56: I liked the blur-effect on the hovering!

AnnikaSonnek (commented):
Good job on this project! I would see if I could add some space between the albums when viewed in mobile or tablet mode and maybe link to the album too! Nicely done!

PR #258, opened 2023-03-10
Eva and Jennifers Music Site

jenster5:
https://eva-and-jennifers-music-site.netlify.app/


code/src/components/Album.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+import { CoverArt } from 'components/CoverArt';
+import { Title } from 'components/Title';
+import { Artists } from 'components/Artists';
+
+export const Album = (props) => {
+  return (
+    <div className="body">
+      <CoverArt link={props.item.external_urls.spotify} images={props.item.images} />
```
khleoth, line 10: Might be something in the name of this reference when you are maping external urls that is not as it should?

code/src/index.css
```diff
@@ -1,13 +1,177 @@
...
+
+.album-image {
+  padding-top: 10px;
+  padding-bottom: 10px;
+  object-fit: cover;
+  width: 100%;
+  height: 100%;
+  transition: filter 0.3s ease-in-out;
+}
+.album-image:hover {
```
khleoth, line 56: I believe  if you put a container that fits both the cover and the buttons, then position the buttons in to stay on top of the image and then add :hover upon that container and display that element on hover you might get sought after result. That is the way I would do it atleast with album cover/icon:hover

khleoth (commented):
First off I want to say good, job, the app seems to display all the necessary data except the artist name not being clickable as far as I can tell on my end and the icons only appear when you hover over them, not the album cover.
Besides that, well done! The site is responsive and seems to be working as intended on the three browsers i tested it on, learning React is tricky, atleast in my opinion :)

PR #261, opened 2023-03-10
Yu Miao & Joanna Philips - Music Releases

Joanna-Philips:
Deployed site: https://music-releases-yu-joanna.netlify.app/


code/src/components/Dropup.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import data from '../stretch-goal.json'
+
+const { useState } = React;
```
jenster5, line 4: impressive you guys did all the stretch goals so cool!

jenster5 (commented):
Wow! your project is great and you did all the stretch goals well done! The code is very clean and easy to read you used many file folders to keep everything neatly organized. Overall just a really well done project.

code/src/components/Dropup.js
EvaSjoholm: I agree with Jennie on this, really nice work!

code/src/components/Album.js
EvaSjoholm, line 1: One Idea could be to use App.js for this section that you've implemented in Album.js, due to also having "Albums.js", which can get a bit confusing (very minor detail). But for separating the app part, and components added. :)

code/src/components/Albums.js
EvaSjoholm, line 1: My App.js, Album.js and Albums.js comment is referring then to this one. Minor details but just a thought.

EvaSjoholm (commented):
Cool that you did stretch-goal and inspiring to see the code. Well done with the design too. Only input would be to use App.js in this specifik so the names for your components are bit clearer for the reader (album.js, albums.js) etc. Well done!

PR #266, opened 2023-03-12
Vera Sjunnesson & Sammy Olsson - Music Releases Project

Vera-Sjunnesson:
View it live:

https://sammy-vera-project-music-releases.netlify.app/


code/src/App.js
```diff
@@ -1,12 +1,20 @@
...
-    <div>
-      Find me in src/app.js!
-    </div>
+    <main>
+      <Header headerName="New Albums" />
+      <AlbumTypeMaster albumTypeCategory="album" />
+      <Header headerName="New Singles" />
+      <AlbumTypeMaster albumTypeCategory="single" />
+      <Header headerName="New Compilations" />
+      <AlbumTypeMaster albumTypeCategory="compilation" />
```
camcron, line 16: Really nice and smart solution to re-use components with different filtering

code/src/components/AlbumTypeMaster.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import data from '../data.json';
+
+import { Album } from './Album';
+
+export const AlbumTypeMaster = (props) => {
+  /* Filtering out albums, singles and compilations depending on the albumCategory value */
+  return (
+    <div className="albums">
```
camcron, line 9: When looking at this div in the chrome inspector tool, we found this className to be a little bit misleading regarding what it contains. Maybe it could be named something like album-wrapper or album-container to make it more obvious that it is a section containing multiple components

code/src/components/AlbumTypeMaster.js
```diff
@@ -0,0 +1,18 @@
...
+      {data.albums.items.filter((item) => item.album_type === props.albumTypeCategory)
+        .map((item) => {
+          return <Album
+            albumCategory={item}
```
camcron, line 13: Not sure of this naming, we understand it gives back a map of all elements in a filtered array based on the category selected, but this make it sound like the props is the category, which we found a little bit confusing when following the code. Could it be named like albumFilteredCategory or filteredAlbum? Not sure ourself, just a thought.

code/src/components/AlbumTypeMaster.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import data from '../data.json';
+
+import { Album } from './Album';
+
+export const AlbumTypeMaster = (props) => {
+  /* Filtering out albums, singles and compilations depending on the albumCategory value */
+  return (
+    <div className="albums">
+      {data.albums.items.filter((item) => item.album_type === props.albumTypeCategory)
```
camcron, line 10: Nicely done with the Conditional (Ternary) Operator

code/src/components/Album.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+import { ArtistName } from './ArtistName';
+import { AlbumCover } from './AlbumCover';
+import { AlbumTitle } from './AlbumTitle';
+
+export const Album = (props) => {
+  return (
+    /* gathering cover, artist name and album title in the same component to later be mapped */
+    <div className="master-wrapper">
```
camcron, line 10: When looking at this div in the chrome inspector tool, we found this className to be a little bit misleading regarding what it contains. Maybe it could be named jus "album" to make it easier to follow.  A general suggestion is to look at the naming of the elements in the inspector tool and try to follow it as if you have never seen this code before.

code/src/components/Album.js
```diff
@@ -0,0 +1,27 @@
...
+        cover={props.albumCategory.images[0].url}
+        link={props.albumCategory.external_urls.spotify} />
+
+      <div className="title-wrapper">
+        <AlbumTitle
+          name={props.albumCategory.name}
+          link={props.albumCategory.external_urls.spotify} />
+
+        <ArtistName
+          artistArray={props.albumCategory.artists} />
```
camcron, line 21: Really good naming, to use "array" in the name makes it so much easier to follow!

code/src/components/Album.js
```diff
@@ -0,0 +1,27 @@
...
+import { ArtistName } from './ArtistName';
+import { AlbumCover } from './AlbumCover';
+import { AlbumTitle } from './AlbumTitle';
+
+export const Album = (props) => {
+  return (
+    /* gathering cover, artist name and album title in the same component to later be mapped */
+    <div className="master-wrapper">
+      <AlbumCover
+        cover={props.albumCategory.images[0].url}
```
camcron, line 12: Nice coding fetching only the first of the image sizes

code/src/components/AlbumCover.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+
+export const AlbumCover = (props) => {
+  return (
+    <div className="album-cover-overlay">
+
+      <a
+        href={props.link}
+        target="_blank"
+        rel="noreferrer"
```
camcron, line 10: Nice touch that you added a link to album from the whole cover image as well as from the album name

code/src/components/AlbumTitle.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+export const AlbumTitle = (props) => {
+  return (
+    <span className="album-name">
```
camcron, line 5: Is there a reason for using a span element to contain the link? If you remove it, you will still only have one jsx-element in the function. We checked the CSS-file and could not find this className being used.

code/src/components/AlbumTitle.js
```diff
@@ -0,0 +1,20 @@
...
+
+      <a
+        href={props.link}
+        target="_blank"
+        rel="noreferrer"
+        className="album-name-link">
+        <h2
+          className="album-name-h2">
```
camcron, line 13: We can see that you have both an "album-name-link" and "album-name-h2". We can't see the reason in the css somewhere. If you want to style them separately you could use this approach since the h2 is inside the a:
.album-link {}
.album-link h2 {}
.album-link a {}

code/src/components/ArtistName.js
```diff
@@ -0,0 +1,24 @@
...
+      {props.artistArray.map((artist) => {
+        return (
+          <a
+            key={artist.id}
+            href={artist.external_urls.spotify}
+            target="_blank"
+            rel="noopener noreferrer">
+            <h3
+              className="artist-name-h3">
+              {artist.name}
```
camcron, line 16: Instead of having a class  on the h3 element called "artist-name-h3", would it be possible to style it like below?
.artist-link h3 {}

code/src/index.css
```diff
@@ -1,13 +1,170 @@
...
 }
+
+main {
+  padding-bottom: 70px;
+}
+
+.albums {
+  display: grid;
+  grid-template-columns: repeat(4, minmax(50px, 1fr));
+  grid-template-rows: repeat(auto, minmax(50px, 1fr));
```
camcron, line 25: Specifying the grid rows is not necessary in this case when the rows have the same size, it will look the same when removing it

code/src/index.css
```diff
@@ -1,13 +1,170 @@
...
+  color: #a0a0a0;
+  display: inline-block;
+}
+
+.artist-name a:not(:first-of-type, :last-of-type)::before {
+  content: ", ";
+}
+
+.artist-name a:nth-last-of-type(2)::after {
+  content: " & ";
```
camcron, line 117: Nicely done adding the “,” and the “&” in the CSS code with pseudo-elements. Did not think of that. Definitely learned something new from this piece of code!

code/src/index.css
```diff
@@ -1,13 +1,170 @@
...
+}
+
+/* HOVER EFFECTS ON ARTIST NAME AND ALBUM NAME */
+.album-name-h2:hover, .artist-name-h3:hover {
+  text-decoration: underline;
+}
+
+/* HOVER EFFECT ON ICON SECTION */
+.album-cover-link:hover + .icons-container {
```
camcron, line 126: Nice use of the adjacent sibling selector!

code/src/index.css
```diff
@@ -1,13 +1,170 @@
...
+  display: flex;
+}
+
+.icons-container:hover {
```
camcron, line 130: We think this is not needed when you have the code on row126? It already adds a display:flex when hovering the image, and you cannot hover the icon-container without also hovering the image.

code/src/index.css
```diff
@@ -1,13 +1,170 @@
...
+.play-icon:hover {
+  transform: scale(1.5);
+  transition: 0.3s ease;
+  opacity: 100%;
+  cursor: pointer;
+}
+
+/* HOVER EFFECTS image */
+.album-cover-overlay:hover .album-cover{
+  opacity: 0.5;
```
camcron, line 152: Minor issue! But it would be clearer to use the same way of specifying opacity, either percentage or a fixed value like 0.5 or 1.

code/src/index.css
```diff
@@ -1,13 +1,170 @@
...
+.album-cover {
+  width: 100%;
+  object-fit: cover;
+}
+
+.album-name-link {
+  display: inline-block;
+}
+
+/* ICONS*/
```
camcron, line 52: Excellent way of working with the icons position and their hover effects. Robust!

camcron (approved):
- Everything looks super nice and functions nice as well, good job!
- Clean code, easy to follow
- Great job on responsiveness
- Very nice solution on the album type sorting & filtering, great to divide it into all components you did.
- One thing that could be improved is using a mobile-first approach in the CSS. 
- Really good job this week! Awesome project! Both of us learned a lot from looking at your code.

// Matilda Frid & Camilla Cronqvist

PR #272, opened 2023-03-12
Emilia, Lina and Andrea music release project

emiliasaberski:


code/src/components/Album.js
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable max-len */
+import React from 'react';
+
+
+
+const Album = (albumTitle) => {
```
fridanordenlow, line 6: Since this component only contains the album title maybe the component could be named AlbumTitle?

code/src/components/AlbumCover.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import Icons from './Icons';
+
+const AlbumCover = (albumCover) => {
+  return (
+    <div className="image-container">
+          <img src={albumCover.cover.images[1].url} className="album-image"/>
+          <Icons />
```
fridanordenlow, line 8: Good solution to create a component with icons to easily import here!

code/src/App.js
```diff
@@ -1,12 +1,28 @@
...
+import AlbumCover from 'components/AlbumCover';
 
-console.log(data);
+import data from './data.json';
 
 export const App = () => {
   return (
     <div>
-      Find me in src/app.js!
+       {data.albums.items.map((album)=> {
```
fridanordenlow: Looking in the inspector tool it looks like since you're doing a map of albums with the css class "container" that is set to "grid", it sets each album to display grid instead of the whole array. If you put a classname to the parent div that returns the map it should work!

code/src/components/Header.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const Header = (props) => {
```
fridanordenlow: The header doesn't appear on the website, looks like it's not mounted anywhere?

fridanordenlow (changes requested):
Great job! :) Just some small styling changes left to fix and you're good to go!

code/src/components/Icons.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import Artist from "./Artist";
+
+const Icons = () => {
+    return (
+    <div className="image-overlay">
+        <div className="icon-container">
```
FionaKlacar, line 7: You have enclosed your icons in two divs here, is there a reason for that?

code/src/components/Icons.js
```diff
@@ -0,0 +1,16 @@
...
+            <img className="icon-dots" src="./icons/dots.svg" />
+            <img className="icon-play" src="./icons/play.svg" />
+            <img className="icon-heart" src="./icons/heart.svg" />
```
FionaKlacar, line 10: Have you thought about adding an alt attribute here?

code/src/index.css
```diff
@@ -5,9 +5,129 @@ body {
...
+
+.albumImage {
+  width: 100%;
+  object-fit: contain;
+  justify-self: center;
+}
+
+.image-overlay {
+  top: 0;
+  z-index: 1;
```
FionaKlacar, line 78: Nice use of the `z-index` here; I guess you were trying to make the album cover fade not obstruct the brightness of the icons? However, I think with` 'image-overlay' `you have targeted only your icons, so you may need to add another z-index to `‘image-container’`.

code/src/index.css
```diff
@@ -5,9 +5,129 @@ body {
...
+  margin: 0 15px;
+}
+
+.icon-play:hover {
+  transform: scale(1.8);
+  transition: 0.3s ease;
+  opacity: 100%;
+}
+
+.icon-heart:hover, .icon-dots:hover {
```
FionaKlacar, line 116: I think only the play button is supposed to get bigger here

FionaKlacar (commented):
Great that you got all the components to work! It will look super once the styling is fixed - well done!

PR #273, opened 2023-03-12
Music project Nina Waldréus

NinaWald:
https://gregarious-malasada-b9b8f8.netlify.app/

### DIFF
--- a/code/src/App.js
+++ b/code/src/App.js
@@ -7,7 +7,7 @@ console.log(data);
 export const App = () => {
   return (
     <div className="main-container">
-      <h1>New albums & singles</h1>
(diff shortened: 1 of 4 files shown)


noraterbocs (commented):
Your code looks quite structured and you have completed all the standard requirements for this assignment. I like it that you found a simple approach to build the project. 
The only thing that I've noticed is that the heart and the dots icons don't have any hover effect. I have seen the code in the css but to be honest I'm not sure why it doesn't work. Also, it would be good to update the readme file as well as it still contains the template. All together  I find your project really well executed, so keep up the good work Nina!

